PR #4, opened 2017-03-03
Optionally run API discovery on service catalog

jwforres:


jwforres:
There is a decent sized whitespace change in openshiftCommon.module.js

jwforres:
So I think the changes to dataService.js and apiService.js might be worth keeping down the road for prototyping against new APIs.  We could keep the getGroups function split out like it is, and then the only thing we would need to take back out later is the actual part that starts the svc cat discovery, and the references to svcCatDeferred.

src/services/apiService.js
```diff
@@ -217,9 +217,11 @@ angular.module('openshiftCommon')
       if (!_.get(APIS_CFG, ["groups", resource.group, "versions", resource.version, "resources", primaryResource])) {
         return undefined;
       }
+      var hostPrefixObj = _.get(APIS_CFG, ["groups", resource.group, 'hostPrefix']) || APIS_CFG;
```
spadgett: Nit: Since `_.get` takes a default, I'd use that.

```js
var hostPrefixObj = _.get(APIS_CFG, ["groups", resource.group, 'hostPrefix'], APIS_CFG);
```

jwforres: That actually only works when get finds an undefined, in this case it gets a null.  I double checked the docs because I was surprised by it.

jwforres:
@jeff-phillips-18 @dtaylor113 to make use of this new optional discovery stuff you need an "additionalServers" section in your config.js, example snippet:

```
window.OPENSHIFT_CONFIG = {
  apis: {
    ...
  },
  api: {
     ...
  },
  additionalServers: [{
    protocol: "http",
    hostPort: "10.245.2.2:30001",
    prefix: "/apis"
  }],

...
```

protocol is optional, it defaults to the protocol you load the page with if not defined. In openshift everything is https but right now svc cat is only http, so this was needed.  For svc cat prefix should always be `/apis`

PR #13, opened 2017-03-22
Add KeywordService, highlightKeywords filter

spadgett:
Pulling the keyword utilities we've written into origin-web-common so they can be used in the catalog.


jwforres:
I know in @dtaylor113 's PR he fixed a typo in the Gruntfile that was affecting dist builds, may need to update this after the PR goes in

https://github.com/openshift/origin-web-common/pull/15/files#diff-35b4a816e0441e6a375cd925af50752cR32

spadgett:
Rebased to pick up the filter fix from @dtaylor113

jwforres:
@jeff-phillips-18 just needs your review

PR #21, opened 2017-03-26
Add `truncate-long-text` directive

spadgett:
Needed to truncate project, service class, and service plan descriptions in origin-web-catalog.


spadgett:
I'm fine changing it, but FWIW, we use kebab case for HTML filenames in origin-web-console. Will push an update.

spadgett:
Updated file names.

PR #28, opened 2017-04-06
Added ui-select.less, bumped version

dtaylor113:


src/styles/_ui-select.less
```diff
@@ -0,0 +1,75 @@
+/* Fix some styles with the angular-ui-select bootstrap theme that isn't right when combined with all of our styles */
+.ui-select-bootstrap {
+  .surface-shaded & {
```
spadgett: This one `.surface-shaded` style should probably stay in origin-web-console since it's specific to a class that's only defined in origin-web-console.

dtaylor113: Ok, removed `.surface-shaded`.

PR #46, opened 2017-04-24
preserve line-breaks in minified file to avoid merge conflicts

liggitt:
copy uglify args from origin-web-console to preserve line breaks in minified output


liggitt:
@spadgett PTAL

spadgett:
Thanks @liggitt

PR #51, opened 2017-04-25
Dedupe kinds returned by APIService.availableKinds()

benjaminapetersen:
Add logic to dedupe kinds by ignoring "openshift" namespace in APIService.calculateAvailableKinds().

Before:
screen shot 2017-04-25 at 3 03 42 pm

After:
screen shot 2017-04-25 at 3 03 25 pm

Updated Fixtures & tests as well.


benjaminapetersen:
@jwforres @spadgett @jeff-phillips-18 @dtaylor113 I assume I should bump the `package.json` and `bower.json` versions, correct? 

 `0.0.17`.

dtaylor113:
> @jwforres @spadgett @jeff-phillips-18 @dtaylor113 I assume I should bump the package.json and bower.json versions, correct?

I believe so, and in the updated origin-web-common dir, execute 'npm publish'

benjaminapetersen:
Special perms for that?  Covers bower as well?

jwforres:
Yes npm publish is a separate list.

spadgett:
I wouldn't worry about the version bump. We can do it as a follow on.

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,50 @@ describe("APIService", function(){
...
+        'Binding', 'ConfigMap', 'Endpoints', 'Event', 'LimitRange', 'PersistentVolumeClaim', 'Pod', 'PodTemplate',
+        'ReplicationController', 'ResourceQuota', 'Secret', 'ServiceAccount', 'Service', 'StatefulSet', 'DeploymentConfig',
+        'LocalSubjectAccessReview', 'LocalResourceAccessReview', 'LocalSubjectAccessReview', 'Policy', 'PolicyBinding',
+        'ResourceAccessReview', 'RoleBindingRestriction', 'RoleBinding', 'Role', 'SelfSubjectRulesReview',
+        'SubjectAccessReview', 'SubjectRulesReview', 'BuildConfig', 'Build', 'DaemonSet', 'Deployment',
+        'HorizontalPodAutoscaler', 'Ingress', 'Job', 'NetworkPolicy', 'ReplicaSet', 'ReplicationControllerDummy',
+        'ImageStreamImage', 'ImageStreamImport', 'ImageStreamMapping', 'ImageStream', 'ImageStreamTag', 'EgressNetworkPolicy',
+        'PodDisruptionBudget', 'AppliedClusterResourceQuota', 'Route', 'PodSecurityPolicyReview',
+        'PodSecurityPolicySelfSubjectReview', 'PodSecurityPolicySubjectReview', 'Template', 'TemplateInstance'
+      ];
```
spadgett: This test is going to break often because the kinds will change. I'd rather not maintain this list if we can avoid it.

Is there another way to test? Maybe check that a subset exists and that OpenShift dupes are not there?

benjaminapetersen: Yeah I'm open to ideas here.  This is not ideal, agree.  Think a subset provides enough assurance?

jwforres:
Yes checking this whole list is probably excessive, but the test is running
off of fixture data not live data @spadgett

test/spec/services/apiServiceSpec.js
spadgett: I'd rather check that a few kinds we know should be there are there, then check that one (or more) cluster-scoped resources we know about **isn't** in that list. We should also test that the OpenShift resources aren't duplicated.

spadgett:
> Yes checking this whole list is probably excessive, but the test is running off of fixture data not live data

Ah then I'm no so concerned

benjaminapetersen:
I pushed up an update with subsets, def agree it will be easier to maintain & prob tests what we need to.

One problem, I'm testing the deduplication, but even ignoring `openshift` there are two kinds with duplication:  `LocalSubjectAccessReview`, and `SubjectAccessReview`.  Thoughts?

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,47 @@ describe("APIService", function(){
...
+    it('should not duplicate kinds that exist in both an api group & the openshift namespace', function() {
+      var allKinds =  _.map(APIService.availableKinds(true), 'kind');
+      var counts = _.reduce(allKinds, function(result, kind) {
+        if(!result[kind]) {
+          result[kind] = 0;
+        }
+        result[kind]++;
+        return result;
+      }, {});
+      // TODO: need a review, not sure why these are still duplicated
```
benjaminapetersen: @spadgett @jwforres need a comment here.  These two still end up with duplication, not sure why. Assuming it is ok?

benjaminapetersen:
@spadgett I did include a bump to the `versions` because I assume I need to update this dependency in `origin-web-console` to finish off the [trello card](https://trello.com/c/EjGHumwk/973-3-handle-effects-of-openshift-api-groups-being-enabled-techdebt).  If that doesn't make sense I can drop it back.

spadgett:
Generally I prefer to make the version bump in a separate commit, but it doesn't matter too much. One of us will need to create a release tag and run `npm publish` when this merges.

benjaminapetersen:
Cool, works for me. easy to roll that back.

jwforres:
Those objects were upstreamed to kube and we kept the same name when we
upstreamed them.  The SubjectAccessReview types don't show up in Other
Resources anyway, they are virtual resources and won't be List-able. They
are technically "different" according to jordan so don't de-dupe them.

benjaminapetersen:
Ok, I'll keep this whitelist in the test then.

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,47 @@ describe("APIService", function(){
...
+
+  describe('#availableKinds', function() {
+    var bothSample = ['Binding','ConfigMap','DeploymentConfig','Event','LimitRange','Pod','ReplicaSet','Role','Service', 'Template'];
+    var onlyClusterSample = ['ClusterResourceQuota','Namespace','OAuthAccessToken','PersistentVolume','ProjectRequest','User'];
+
+    // at minimum, prove with tests that it is deduplicating by ignoring openshift!
+    it('should return list of kinds that are scoped to a namespace by default', function() {
+      var namespacedKinds = _.map(APIService.availableKinds(), 'kind');
+      // from the fixtures included, we should have 51 namespaced kinds
+      expect(namespacedKinds.length).toEqual(51);
```
jwforres: eh even using fixture data i dont love this, i'd rather do something like validate the list does not include a known cluster scoped kind

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,47 @@ describe("APIService", function(){
...
+      expect( _.difference(bothSample, namespacedKinds).length ).toEqual(0);
+    });
+
+    it('should return list of all kinds, including those that are cluster scoped, when passed a truthy argument', function() {
+      var allKinds = _.map(APIService.availableKinds(true), 'kind');
+      // from the fixtures included, we should have 84 total kinds
+      expect(allKinds.length).toEqual(84);
```
jwforres: take this out

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,47 @@ describe("APIService", function(){
...
+      expect( _.difference(bothSample, allKinds).length ).toEqual(0);
+      expect( _.difference(onlyClusterSample, allKinds).length ).toEqual(0);
+    });
+
+    it('should not duplicate kinds that exist in both an api group & the openshift namespace', function() {
```
jwforres: your test is not actually validating this statement, your test is validating there are no duplicates at all (based on current fixture data) except for ones you explicitly allow

i see benefit in having both tests, but they are distinct

test/spec/services/apiServiceSpec.js
benjaminapetersen: Agree, that makes more sense. I'll update

test/spec/services/apiServiceSpec.js
benjaminapetersen: will do

test/spec/services/apiServiceSpec.js
benjaminapetersen: Ah, yeah I conflated those into one, will update.

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,96 @@ describe("APIService", function(){
...
+      expect( _.difference(bothSample, namespacedKinds).length ).toEqual(0);
+    });
+
+    it('should return list of all kinds, including those that are cluster scoped, when passed a truthy argument', function() {
+      var allKinds = _.map(APIService.availableKinds(true), 'kind');
+      // from the fixtures included, we should have 84 total kinds
+      expect( _.difference(bothSample, allKinds).length ).toEqual(0);
+      expect( _.difference(onlyClusterSample, allKinds).length ).toEqual(0);
+    });
+
```
benjaminapetersen: @jwforres I think the following 3 tests are now accurate.  Thoughts?

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,96 @@ describe("APIService", function(){
...
+
+  describe('#availableKinds', function() {
+    var bothSample = ['Binding','ConfigMap','DeploymentConfig','Event','LimitRange','Pod','ReplicaSet','Role','Service', 'Template'];
+    var onlyClusterSample = ['ClusterResourceQuota','Namespace','OAuthAccessToken','PersistentVolume','ProjectRequest','User'];
+
+    // at minimum, prove with tests that it is deduplicating by ignoring openshift!
+    it('should return list of kinds that are scoped to a namespace by default', function() {
+      var namespacedKinds = _.map(APIService.availableKinds(), 'kind');
+      // from the fixtures included, we should have 51 namespaced kinds
```
jwforres: can you update the comments, they are still based on your old version of the test

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,96 @@ describe("APIService", function(){
...
+      expect( _.difference(bothSample, namespacedKinds).length ).toEqual(0);
```
jwforres: this doesnt validate that cluster kinds are not included, you need two expectations.  So you need a second line here that checks the difference between cluster sample and all kinds is not 0

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,96 @@ describe("APIService", function(){
...
+      ];
+      _.each(k8sAPIStillExistsSample, function(kindToFind) {
+        var found = _.find(allKinds, function(kind) {
+          return (kind.kind === kindToFind.kind) && !_.includes(_.keys(kind), 'group');
+        });
+        if(found) {
+          shouldBeFound.push(found);
+        }
+      });
+      expect(shouldBeFound.length).toEqual(7);
```
jwforres: instead of hardcoding 7 here, the test-case would be more future-proof if you did k8sAPIStillExistsSample.length  that way if someone updates the list they dont need to remember to update the number here

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,96 @@ describe("APIService", function(){
...
+    });
+
+    // When a kind is promoted from openshift to kube, for a time, we will allow it to exist in both api groups
+    it('should not return duplicate kinds, except for those we have promoted from openshift to kube', function() {
```
jwforres: i would eliminate this test altogether, its just not true.  there may always be duplicate kinds when we have aggregated api servers doing any number of things, i think your other two tests confirm the right thing, which is that we no longer include oapi in the list of available kinds, but that we do still include /api

benjaminapetersen: ok dropping this one.

test/spec/services/apiServiceSpec.js
benjaminapetersen: agree, updating

test/spec/services/apiServiceSpec.js
benjaminapetersen: agree! updating.

benjaminapetersen:
updated

test/spec/services/apiServiceSpec.js
```diff
@@ -234,6 +235,78 @@ describe("APIService", function(){
...
+    var onlyClusterSample = ['ClusterResourceQuota','Namespace','OAuthAccessToken','PersistentVolume','ProjectRequest','User'];
+
+    it('should return list of kinds that are scoped to a namespace by default', function() {
+      var namespacedKinds = _.map(APIService.availableKinds(), 'kind');
+      expect( _.difference(bothSample, namespacedKinds).length ).toEqual(0);
+    });
+
+    it('should not return list cluster scoped kinds by default', function() {
+      var namespacedKinds = _.map(APIService.availableKinds(), 'kind');
+      expect( _.difference(onlyClusterSample, namespacedKinds).length ).toEqual(6);
```
jwforres: dont hardcode this to '6', set it to onlyClusterSample.length

benjaminapetersen: ah, agree. oop.

PR #55, opened 2017-05-02
Adding shared mixins and variables

rhamilto:
Fixes https://github.com/openshift/origin-web-console/issues/1500

Once this merges and a new release is cut:

* I will open a PR that utilizes the new release in origin-web-catalog

Once that aforementioned PR to origin-web-catalog merges and a new release of it is cut:

* I will open a PR to origin-web-console that utilizes the new releases of origin-web-common and origin-web-catalog


spadgett:
LGTM, but we let's cut a release first with just #51

PR #60, opened 2017-05-05
Always toast project delete

### LINKED ISSUE #59: delete-project component shows both toast and in-page alert
If you pass `alerts` to the `delete-project` component, it shows both a toast and an in-page alert.

screen shot 2017-05-05 at 8 41 05 am

If you don't pass alerts, it results in a runtime error.

It makes sense to me to use the alerts when they're passed in and use toasts when they aren't.

openshift web console 2017-05-05 08-38-33

@jwforres @jeff-phillips-18

spadgett:
Use toast notifications always instead of inline alerts when users delete projects.

Fixes #59

@jeff-phillips-18 @jwforres @serenamarie125 any objections to this change?

### DIFF
--- a/dist/origin-web-common-ui.js
+++ b/dist/origin-web-common-ui.js
@@ -336,14 +336,12 @@ angular.module("openshiftCommonUI")
 ;'use strict';
 
 angular.module("openshiftCommonUI")
-  .directive("deleteProject", function ($uibModal, $location, $filter, $q, hashSizeFilter, APIService, DataService, AlertMessageService, NotificationsService, Logger) {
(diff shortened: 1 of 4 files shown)


jwforres:
it makes it consistent with project delete when the new experience is on, and inconsistent with everywhere else we are still doing inline alerts.  if we are going to make the switch i'd prefer to see us make the switch everywhere as a card so QE can validate it and fix their automation.  is there any argument against switching all of our inline success alerts to be toasts instead?

jwforres:
we probably shouldn't switch our existing inline warning/alert errors though, these i think are generally what we want in the places they are being used (like after form submission fails)

spadgett:
We're already inconsistent since we are showing toasts for create project, but nothing else :/

I think we should bite the bullet and just update what we think make sense (delete, successful edits, maybe start deployment / build, etc.)

PR #61, opened 2017-05-05
Update calculateAvailableKinds to support both string & object

benjaminapetersen:


dist/origin-web-common-services.js
```diff
@@ -478,7 +483,7 @@ angular.module('openshiftCommonServices')
       _.each(api.resources.v1, function(resource) {
         if (resource.namespaced || includeClusterScoped) {
           // Exclude subresources and any rejected kinds
-          if (resource.name.indexOf("/") >= 0 || _.contains(rejectedKinds, resource.kind)) {
+          if (_.contains(resource.name, '/') || _.find(rejectedKinds, {kind: resource.kind})) {
```
jwforres: dont you need `group: ''` here

dist/origin-web-common-services.js
```diff
@@ -495,7 +500,7 @@ angular.module('openshiftCommonServices')
       var preferredVersion = defaultVersion[group.name] || group.preferredVersion;
       _.each(group.versions[preferredVersion].resources, function(resource) {
         // Exclude subresources and any rejected kinds
-        if (resource.name.indexOf("/") >= 0 || _.contains(rejectedKinds, resource.kind)) {
+        if (_.contains(resource.name, '/') || _.find(rejectedKinds, {kind: resource.kind})) {
```
jwforres: you need group here, or its just going to do the same thing it does today

benjaminapetersen: Tinkering with this, atm it is greedy, dropping both `Binding` and `{ kind: "Binding", group: "servicecatalog.k8s.io" }`.

dist/origin-web-common.js
```diff
@@ -896,44 +896,6 @@ angular.module('openshiftCommonUI')
       }
     };
   }]);
```
benjaminapetersen: Pulling this out into the `test/fixtures` dir.... was it supposed to end up in the dist?

jwforres: from the perspective of what should happen in the console, we never want OPENSHIFT_CONFIG being included in dist, that file should always be guaranteed to come from the generated config.js and nowhere else.  So removing it from common's dist is correct.  If that breaks the catalog app for some reason then we'll need to fix the catalog app.

benjaminapetersen: Yeah that seemed off.

benjaminapetersen:
Ok, I think this is ready for review.  Added a test & pulled some things out into fixtures.

test/spec/fixtures/constants.js
```diff
@@ -0,0 +1,12 @@
+window.OPENSHIFT_CONSTANTS = window.OPENSHIFT_CONSTANTS || {};
+
+window.OPENSHIFT_CONSTANTS.AVAILABLE_KINDS_BLACKLIST = [
+    // These are k8s kinds that are not supported in the current release of OpenShift
+    { kind: "Binding", group: "servicecatalog.k8s.io" },
```
jwforres: Update the comment so its clear that its only being blacklisted for testing purposes, the comment as-is is definitely misleading, we won't be blacklisting the `servicecatalog.k8s.io` Binding in the console.

benjaminapetersen: will do.

PR #62, opened 2017-05-05
Use Patternfly toast notifications for DataService errors

spadgett:
Use Patternfly toast notifications for failed list, get, and watch requests. This removes our dependency on messenger. It also removes the `Notification` service, whose name is confusingly similar to `NotificationService`.

This change also prevents permanently-hidden notification from appearing in the drawer (when we have a drawer).

screen shot 2017-05-05 at 2 24 29 pm


src/ui-services/notificationsService.js
```diff
@@ -17,9 +17,9 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
       return 'hide/notification/' + namespace + '/' + notificationID;
     };
 
-    var addNotification = function (notification, notificationID, namespace) {
-      if (notificationID && isNotificationPermanentlyHidden(notificationID, namespace)) {
-        notification.hidden = true;
+    var addNotification = function (notification) {
```
spadgett: @jeff-phillips-18 I changed the signature here because I need to be able to get the ID later for `isNotificationVisible`. I don't think we were using this anywhere with a `notificationID` yet, correct?

jwforres:
LGTM, will wait for @jeff-phillips-18 's answer about notificationID

src/ui-services/notificationsService.js
jeff-phillips-18: @spadgett correct, this is not yet being used.

PR #63, opened 2017-05-09
Incorporate slide in/out animation for toast notifications

sg00dwin:
- multiple toasts stack with 1px margin between
- multiple toasts have matching widths
- moved top alignment beneath top thin bar

screen shot 2017-05-08 at 2 26 12 pm
screen shot 2017-05-08 at 2 08 29 pm

### DIFF
--- a/dist/less/_animations.less
+++ b/dist/less/_animations.less
@@ -1,0 +1,21 @@
+// Toast notification
+@keyframes toastSlideIn {
+  0% {
+    opacity: 0;
+    transform: translateX(150px);
+  }
+  100% {
+    opacity: 1;
+    transform: translateX(0);
+  }
+}
+@keyframes toastSlideOut {
+  0% {
+    opacity: .75;
+    transform: translateX(0);
+  }
+  100% {
+    opacity: 0;
+    transform: translateX(150px);
+  }
+}
--- a/dist/less/_notifications.less
+++ b/dist/less/_notifications.less
@@ -6,3 +6,20 @@
   font-weight: 700;
   margin-right: 5px;
 }
+
+.toast-notifications-list-pf {
+  top: 28px;
+  > div {
+    &.ng-enter {
+      animation: toastSlideIn .2s ease-out;
+    }
+    &.ng-leave {
+      animation: toastSlideOut .2s ease-in;
+    }
+    .toast-pf {
+      border-color: @color-pf-white;
+      margin-bottom: 5px;
+      width: 100%;
+    }
+  }
+}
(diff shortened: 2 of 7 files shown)


spadgett:
@sg00dwin 1px might be a little tight?

@serenamarie125 Wanted to get your input on this change. This is what @sg00dwin is addressing. The different widths look odd and the background page content visible between the alerts is distracting to me.

screen shot 2017-05-05 at 2 20 18 pm

cc @jwforres @rhamilto

spadgett:
cc @jeff-phillips-18

cshinn:
So existing toasts will lengthen to match the size of the largest one? There isn't necessarily anything wrong with that, but it may lead to a large toast wall if there are several short ones and a long message. Is there a maximum width for these toasts?

serenamarie125:
I agree that the spacing between toasts is way too much, but i think that the proposed 1px is too little.  I would suggest somewhere between 2 - 5 px between them @beanh66.  We could do a mockup, or if you guys can try it quickly and add screenies that would work as well.  Let us know! @spadgett @sg00dwin

spadgett:
@cshinn I believe they're a fixed width, @sg00dwin can you confirm?

spadgett:
@serenamarie125 Thanks! We'll get some screenshots.

We could contribute the animation back upstream, although currently it is specific to an angular implementation using `ng-repeat`

spadgett:
@serenamarie125 @beanh66 

5px margin:

openshift web console 2017-05-10 10-42-45

2px margin:

openshift web console 2017-05-10 10-43-41

Let me know if you want 3 or 4 px

spadgett:
I found a style in origin-web-console for inline alerts that unintentionally makes the toast margin wider than it should be (as in my first screenshot above of the problem). Patternfly defaults appear to be 14px, which looks like:

openshift web console 2017-05-10 10-46-23

beanh66:
@spadgett Thanks for sharing, I like 5px better than 2. @serenamarie125 thoughts?

serenamarie125:
@spadgett thanks!  I'd move ahead with 5px.  In the meantime, I'll create an issue in PF to see if they can fix it in the interim.  Once that occurs, we should go back to PF.

sg00dwin:
@cshinn @spadgett PatternFly sets a max-width on toasts when the browsers is > 768px

serenamarie125:
fyi @spadgett https://github.com/patternfly/patternfly/issues/642

sg00dwin:
@spadgett updated pr to 5px `margin-bottom`

screen shot 2017-05-10 at 1 16 55 pm

PR #68, opened 2017-05-15
Clarifying flex use case for word-break-all

rhamilto:


sg00dwin:
@rhamilto looks good

spadgett:
Wonder if we shouldn't add `word-break-flex` class that adds the `flex: 1 1 0%`?

rhamilto:
> Wonder if we shouldn't add word-break-flex class that adds the flex: 1 1 0%?

What do you think, @sg00dwin?  I have mixed feelings...

sg00dwin:
I'm not sure what having a separate `word-break-flex` gets us. Wouldn't that lead to having other cases `word-break-table`, etc.

spadgett:
I'm OK not doing it. My only thought is that the `flex: 1 1 0%` is not at all obvious, and we'll need to comment everywhere we use it that it's required for word-break

rhamilto:
I can haz merge?

PR #71, opened 2017-05-16
Move ResourceGroupVersion constructor to service to eliminate global

benjaminapetersen:
This just wraps the constructor function in a service so we eliminate the global.

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -203,45 +203,48 @@ angular.module("openshiftCommonServices")
   });
 ;'use strict';
 
-// ResourceGroupVersion represents a fully qualified resource
-function ResourceGroupVersion(resource, group, version) {
-  this.resource = resource;
-  this.group    = group;
-  this.version  = version;
-  return this;
-}
-// toString() includes the group and version information if present
-ResourceGroupVersion.prototype.toString = function() {
-  var s = this.resource;
-  if (this.group)   { s += "/" + this.group;   }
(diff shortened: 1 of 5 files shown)


spadgett:
Aren't there many instances of ResourceGroupVersion whereas a service is a singleton?

jwforres:
yeah agree with @spadgett here, this change does not look correct

benjaminapetersen:
Correct, you still call: 
```JavaScript 
new ResourceGroupVersion(resource, group, version);
``` 
this just puts the constructor inside a service, rather than making it a global that attaches to `window` as `window.ResourceGroupVersion`. 

Alternatively, I could move lines `3-39` inside the `.factory('APIService')`, it makes little difference to me (just eliminating the global).

jwforres:
I see what you are saying now, and I definitely do not like defining a "Service" that isn't truly a service. I'd rather see the global moved underneath the service its being used by which is APIService, we should never be directly creating ResourceGroupVersion objects without using APIService so that we guarantee proper resolutions are happening.

benjaminapetersen:
Ok, I may just wrapping the whole file in:

```JavaScript 
(function() {
 // code here, yay
})();
```
& that prob will do the trick w/no real changes.

benjaminapetersen:
Updated & just wrapped in `(function() { })()`.
Also updated `kindToResouce` to change param from `kind` to `kindStr`, just to clarify it doesn't take a kind object (ran into that issue).  This is not a real change so much as self documenting.

PR #72, opened 2017-05-17
Adding line-clamp mixin

rhamilto:
The mixin was created as part of https://github.com/openshift/origin-web-catalog/pull/230, but it should really live in common so that it can be used in both the console and the catalog.  I will update https://github.com/openshift/origin-web-catalog/pull/230 to remove this mixin from that pr once this PR lands and catalog can be updated to make use of it.


spadgett:
@rhamilto OK to merge when travis passes

PR #77, opened 2017-05-23
Update APIService.apiInfo to return verbs when possible

benjaminapetersen:
For last task on [this trello card](https://trello.com/c/EjGHumwk/973-3-handle-effects-of-openshift-api-groups-being-enabled-techdebt).
PR to origin-web-console coming soon depending on this change.

### DIFF
--- a/src/services/apiService.js
+++ b/src/services/apiService.js
@@ -188,6 +188,13 @@ angular.module('openshiftCommonServices')
     return resource;
   }
 
+  function kindToResourceGroupVersion(kind) {
+    return toResourceGroupVersion({
+      resource: kindToResource(kind.kind),
+      group: kind.group
+    });
+  }
+
   // apiInfo returns the host/port, prefix, group, and version for the given resource,
   // or undefined if the specified resource/group/version is known not to exist.
   var apiInfo = function(resource) {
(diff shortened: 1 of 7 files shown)


src/services/apiService.js
```diff
@@ -239,9 +248,10 @@ angular.module('openshiftCommonServices')
         continue;
       }
       return {
+        resource: resource.resource,
+        version:  resource.version,
         hostPort: api.hostPort,
         prefix:   api.prefix,
-        version:  resource.version,
         namespaced: discoveredResource.namespaced
```
jwforres: why arent you getting the verbs here as well, im looking at the api discovery on my machine and i get verbs on these

src/services/apiService.js
```diff
@@ -188,6 +187,14 @@ angular.module('openshiftCommonServices')
     return resource;
   }
 
+  function kindToResourceGroupVersion(kind) {
+    var resourceStr = kindToResource(kind.kind);
```
spadgett: I'm confused by this. I'd expcct this to take in a string like `"ReplicaSet"` rather than an object. We might think about the function and arg name here.

src/services/apiService.js
benjaminapetersen: I was not getting verbs, but let me check, maybe i wan't on latest origin.

benjaminapetersen: Yup, i get them now. ok will update., my bad.

benjaminapetersen:
Updated, added tests & updated fixtures (& fixed a broken DataService test)

benjaminapetersen:
[test]

PR #79, opened 2017-05-23
Drop PhantomJS & use Firefox for tests

benjaminapetersen:
Dropping PhantomJS & migrate to FF like we did in web-console.

@jwforres @jeff-phillips-18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ cache:
   directories:
   - node_modules
   - bower_components
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3 # give xvfb some time to start
 before_script:
   - make build
 script:
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -52,7 +52,10 @@ module.exports = function (grunt) {
         unit: {
           configFile: 'test/karma.conf.js',
           singleRun: true,
-          browsers: ['PhantomJS']
+          // grunt test --browsers=Chrome,Firefox,Safari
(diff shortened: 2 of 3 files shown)


jwforres:
im good with the change in theory but looks like its failing to start firefox in travis

benjaminapetersen:
I'll look at the `travis.yaml` here in just a sec. might be missing something.

benjaminapetersen:
[TEST] `travis.yml` updated

benjaminapetersen:
Think we good here now.  Needed the `xvfb` install stuff.

.travis.yml
```diff
@@ -5,6 +5,10 @@ cache:
   directories:
   - node_modules
   - bower_components
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
```
jwforres: test-headless should be doing this for you already https://github.com/openshift/origin-web-common/blob/master/hack/test-headless.sh#L34

and if you look at our travis.yml in origin-web-console this wasn't necessary

benjaminapetersen: yeah thought so, but this seemed to make it work (is what i copied from other feeder repos like kve & extension-registry).  Ill dig around a bit more.

spadgett:
Recently Travis updates broke origin-web-console tests using Firefox. We probably need to hold off on this.

benjaminapetersen:
Adding the comment/links here as well, [nightmare](https://github.com/segmentio/nightmare) [nightmare](http://www.nightmarejs.org/) could be worth a shot.  I may tinker with it a bit & see if there is any benefit.

PR #89, opened 2017-05-31
Adding HTMLService and linkify filter from console

rhamilto:
@sg00dwin and I need to make use of HTMLService.getBreakpoint() on catalog landing page, so I'm migrating HTMLService and related linkify filter to common.

### DIFF
--- a/src/openshiftCommonUI.module.js
+++ b/src/openshiftCommonUI.module.js
@@ -5,6 +5,15 @@
  *   Base module for openshiftCommonUI.
  */
 angular.module('openshiftCommonUI', [])
+// Sometimes we need to know the css breakpoints, make sure to update this
+// if they ever change!
+.constant("BREAKPOINTS", {
+  screenXsMin:  480,   // screen-xs
+  screenSmMin:  768,   // screen-sm
+  screenMdMin:  992,   // screen-md
+  screenLgMin:  1200,  // screen-lg
+  screenXlgMin: 1600   // screen-xlg
+})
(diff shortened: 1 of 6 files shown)


src/openshiftCommonUI.module.js
```diff
@@ -5,6 +5,15 @@
...
 angular.module('openshiftCommonUI', [])
+// Sometimes we need to know the css breakpoints, make sure to update this
+// if they ever change!
+.constant("BREAKPOINTS", {
+  screenXsMin:  480,   // screen-xs
+  screenSmMin:  768,   // screen-sm
+  screenMdMin:  992,   // screen-md
+  screenLgMin:  1200,  // screen-lg
+  screenXlgMin: 1600   // screen-xlg
+})
```
jeff-phillips-18: Hmm... Some if this is defined in patternfly:

  patternfly.pfBreakpoints = {
    'tablet': 768,
    'desktop': 1200
  };

Maybe the other values ought to be defined there as well. 
Until then, the patternfly constant is used in the catalog repo already(https://github.com/openshift/origin-web-catalog/blob/master/src/components/saas-list/saas-list.controller.ts#L51-L52). We may want to update that to use this constant instead.

rhamilto: Good call, @jeff-phillips-18.  Changed in https://github.com/openshift/origin-web-catalog/pull/261/commits/a08c2a4f12925a0f4a37cbbe8602cb794ff4a04b#diff-052c10f2f663033239b7d443fe45f0eaR52 (along with SaaS mobile styling changes).

PR #99, opened 2017-06-08
Optionally send AlphaPodPresetTemplate when creating Binding

jwforres:


jwforres:
@spadgett this is probably all we need for common, paul and I are debugging through an issue atm though so hold off on merging

PR #101, opened 2017-06-09
Migrating popup directive from console

rhamilto:
@sg00dwin needs popups on the project list on the catalog landing page, so I'm migrating the popup directive to common so it can be used there as well as the console.

### DIFF
--- a/src/components/popups/popups.js
+++ b/src/components/popups/popups.js
@@ -1,0 +1,88 @@
+'use strict';
+
+angular.module('openshiftCommonUI')
+  // This triggers when an element has either a toggle or data-toggle attribute set on it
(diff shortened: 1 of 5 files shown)


src/components/popups/popups.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+angular.module('openshiftCommonUI')
+  // This triggers when an element has either a toggle or data-toggle attribute set on it
+  .directive('toggle', function(IS_IOS) {
```
jeff-phillips-18: The IS_IOS constant be added to the CommonUI module rather than depending on the applications to set it.

rhamilto: I've moved IS_IOS to common in this PR.

PR #104, opened 2017-06-12
[WIP] Remove unused AlertMessageService methods

spadgett:
Remove addAlert(), getAlerts(), and clearAlerts(). These are no longer
used now that we use toast notifications for showing alerts just before
a page transition.

See https://github.com/openshift/origin-web-console/pull/1679
See https://github.com/openshift/origin-web-catalog/pull/278


spadgett:
Marking WIP until https://github.com/openshift/origin-web-console/pull/1679 merges

PR #107, opened 2017-06-13
Truncate Update

### LINKED ISSUE #98: truncate when exact character count results in ellipsis and "See All" even though there is nothing more
Character count is exact match for truncate, so . is removed, ... and See All is added.
screen shot 2017-06-08 at 9 37 38 am

Clicking "See All" restores the .
screen shot 2017-06-08 at 9 37 43 am

cdcabrera:
Issue fix, return string when matching character limit param is met.

### DIFF
--- a/src/filters/truncate.js
+++ b/src/filters/truncate.js
@@ -3,7 +3,7 @@
 angular.module('openshiftCommonUI')
   .filter('truncate', function() {
     return function(str, charLimit, useWordBoundary, newlineLimit) {
-      if (!str) {
+      if (typeof str !== 'string') {
         return str;
       }
 
@@ -19,11 +19,10 @@ angular.module('openshiftCommonUI')
       }
 
       if (useWordBoundary !== false) {
-        // Find the last word break, but don't look more than 10 characters back.
-        // Make sure we show at least the first 5 characters.
+
         var startIndex = Math.max(4, charLimit - 10);
-        var lastSpace = truncated.lastIndexOf(/\s/, startIndex);
-        if (lastSpace !== -1) {
+        var lastSpace = truncated.lastIndexOf(' ');
+        if (lastSpace >= startIndex && lastSpace !== -1) {
           truncated = truncated.substring(0, lastSpace);
         }
       }
--- a/test/spec/filters/truncateSpec.js
+++ b/test/spec/filters/truncateSpec.js
@@ -1,0 +1,97 @@
+describe("TruncateFilter", function() {
+
+  var $scope;
+  var $filter;
+
+  beforeEach(inject(function (_$rootScope_, _$filter_) {
+
+    $scope = _$rootScope_;
+    $filter = _$filter_;
+  }));
+
+  beforeEach(function () {
+
+    $scope.params = {
+      str: 'Operation cannot be fulfilled on namespaces\n"ups-broker": The system is ensuring all content is removed\nfrom this namespace. Upon completion, this\nnamespace will automatically be purged by the system.\n',
+      charLimit: 200,
+      useWordBoundary: false,
+      newlineLimit: null
+    };
+  });
+
+  it('should have a minimum of two parameters defined', function() {
+
+    var str = null;
+    var charLimit = null;
+    var useWordBoundary = $scope.params.useWordBoundary;
+    var newlineLimit = $scope.params.newlineLimit;
+    var filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+
+    expect(str).toBeFalsy();
+    expect(charLimit).toBeFalsy();
+    expect(useWordBoundary).toEqual(false);
+    expect(newlineLimit).toBeFalsy();
+
+    expect(filteredString).toEqual(null);
+  });
+
+  it('should return a string with a specific length', function() {
+
+    var str = $scope.params.str;
+    var charLimit = $scope.params.charLimit;
+    var useWordBoundary = $scope.params.useWordBoundary;
+    var newlineLimit = $scope.params.newlineLimit;
+    var filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+
+    expect(filteredString.length).toEqual(200);
+  });
+
+  it('should return a string that ends on a word boundary', function() {
+
+    var str = $scope.params.str;
+    var charLimit = $scope.params.charLimit;
+    var useWordBoundary = true;
+    var newlineLimit = $scope.params.newlineLimit;
+    var filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+
+    expect(filteredString.length).toEqual(192);
+    expect(/^\s/.test(str.split(filteredString)[1])).toEqual(true);
+  });
+
+  it('should return a string with a limited number of new lines', function() {
+
+    var str = $scope.params.str;
+    var charLimit = $scope.params.charLimit;
+    var useWordBoundary = $scope.params.useWordBoundary;
+    var newlineLimit = 1;
+    var filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+
+    expect((filteredString.match(/\n/g)||[]).length).toEqual(0);
+
+    newlineLimit = 2;
+    filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+    expect((filteredString.match(/\n/g)||[]).length).toEqual(1);
+
+    newlineLimit = 3;
+    filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+    expect((filteredString.match(/\n/g)||[]).length).toEqual(2);
+
+    newlineLimit = 4;
+    filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+    expect((filteredString.match(/\n/g)||[]).length).toEqual(3);
+  });
+
+  it('should return a string that both ends on a word boundary, and has a limited number of new lines', function() {
+
+    var str = $scope.params.str;
+    var charLimit = $scope.params.charLimit;
+    var useWordBoundary = true;
+    var newlineLimit = 3;
+    var filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+
+    expect(filteredString.length).toEqual(146);
+    expect(/^\s/.test(str.split(filteredString)[1])).toEqual(true);
+    expect((filteredString.match(/\n/g)||[]).length).toEqual(2);
+  });
+
+});
--- a/dist/origin-web-common-ui.js
+++ b/dist/origin-web-common-ui.js
@@ -1682,7 +1682,7 @@ angular.module('openshiftCommonUI')
 angular.module('openshiftCommonUI')
   .filter('truncate', function() {
     return function(str, charLimit, useWordBoundary, newlineLimit) {
-      if (!str) {
+      if (typeof str !== 'string') {
         return str;
       }
 
@@ -1698,11 +1698,10 @@ angular.module('openshiftCommonUI')
       }
 
       if (useWordBoundary !== false) {
-        // Find the last word break, but don't look more than 10 characters back.
-        // Make sure we show at least the first 5 characters.
+
         var startIndex = Math.max(4, charLimit - 10);
-        var lastSpace = truncated.lastIndexOf(/\s/, startIndex);
-        if (lastSpace !== -1) {
+        var lastSpace = truncated.lastIndexOf(' ');
+        if (lastSpace >= startIndex && lastSpace !== -1) {
           truncated = truncated.substring(0, lastSpace);
         }
       }
--- a/dist/origin-web-common.js
+++ b/dist/origin-web-common.js
@@ -1853,7 +1853,7 @@ angular.module('openshiftCommonUI')
 angular.module('openshiftCommonUI')
   .filter('truncate', function() {
     return function(str, charLimit, useWordBoundary, newlineLimit) {
-      if (!str) {
+      if (typeof str !== 'string') {
         return str;
       }
 
@@ -1869,11 +1869,10 @@ angular.module('openshiftCommonUI')
       }
 
       if (useWordBoundary !== false) {
-        // Find the last word break, but don't look more than 10 characters back.
-        // Make sure we show at least the first 5 characters.
+
         var startIndex = Math.max(4, charLimit - 10);
-        var lastSpace = truncated.lastIndexOf(/\s/, startIndex);
-        if (lastSpace !== -1) {
+        var lastSpace = truncated.lastIndexOf(' ');
+        if (lastSpace >= startIndex && lastSpace !== -1) {
           truncated = truncated.substring(0, lastSpace);
         }
       }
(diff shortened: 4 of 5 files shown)


cdcabrera:
Appears like there could be additional conflicts in the truncate sequence, where
- a hard truncate fires first
- new line truncate fires second
- word boundary truncate fires last

A potential resequence could look like:
- new line truncate firing first
- and an else on hard and word boundary truncates

src/filters/truncate.js
```diff
@@ -3,7 +3,7 @@
 angular.module('openshiftCommonUI')
   .filter('truncate', function() {
     return function(str, charLimit, useWordBoundary, newlineLimit) {
-      if (!str) {
+      if (typeof str !== 'string' || str.length <= charLimit) {
```
spadgett: If there is a hard newline limit like 1 line, this will return a string with newlines. Can we apply `useWordBoundary` only if the string is longer than limit, but still apply any newline truncation below?

cdcabrera: As a long term fix, indeed we could look at a re-sequence, the immediate issue appears resolved.

The potential conflict mentioned above appeared after running a hard truncate then applying newline truncate. If the newline truncate is meant to preserve the height of a container, then running that first in sequence appears to reduce the greater number of characters first. After that, as your question mentions, running a word boundary, or even a hard character limit, would help narrow it down further. 

Hard truncate and word boundary slightly boil down to interpretation, the sequence that currently exists, from observation, appears geared toward preserving enough space for a link, which in a way, slightly feels like the responsibility of the parent not the filter itself.

spadgett: My complaint is that it changes the behavior of `"foo\nbar\n"` where length limit is 80 and newline limit is 1. Before this would remove the second line and now it will return the string as-is. I think we can fix the immediate bug while still keeping the newline limit without too many other changes.

But I think the problem might be different here: If truncate is 80 characters and the string is 81, we're taking off one character and appending and ellipsis, which actually makes the string longer. Or is there another issue at play?

cdcabrera: Gotcha, making an update now for a quick check on newlines prior to returning.

Regarding the second part about adding/subtracting characters... going to prefix it with a question. 
- Are we aiming to do whole word truncates, where it properly rounds a word to completion, or do we just want to cut the string? 

If we're going for a ```charLimit truncate``` as an end result we're good to go. If we want to end on a word boundary when ```useWordBoundary``` is active then we may need to modify the use of [```charLimit - 10``` on line 24](https://github.com/openshift/origin-web-common/blob/master/src/filters/truncate.js#L24)

spadgett: I think we both want to make sure the string isn't longer than limit and truncate on word boundary when we can.

The intent there is that we will not look back farther than 10 characters to find a word boundary in case we're truncating in the middle of a long unbroken string like a sha identifier or uid.

spadgett:
So I think there are two problems here:

1. We're still trying to apply word boundaries when no truncation was needed. @cdcabrera Have you confirmed this is definitely a problem?
2. We sometimes truncate and make the string longer. (We chop one character off and then add a three-character ellipses.)

I think the change addresses (1) but not (2). For (1), I would suggest something like

```js
      if (charLimit) {
        truncated = truncated.substring(0, charLimit);
        if (str.length === truncated.length) {
          // Only check for word boundaries if we're truncating based on length and not newlines.
          useWordBoundary = false;
        }
      }
```

For (2), we might want an additional option to say only truncate if >n chars over the limit. For instance, only truncate to 80 if it is at least 83 chars long because I am going to add an ellipsis. (The  `truncate-long-text` directive adds an ellipsis.)

cdcabrera:
@spadgett 

Regarding ```1``` indeed, yes we're still trying to continue the process when it's unnecessary to truncate. 

For ```2``` agreed... 
- we're removing characters so we can add characters for sure, yes, which is causing the partial "hard" truncate on the original issue where the period is removed. And 
- we probably need to either pass a parameter to the filter that removes said characters at the start of the process. 

And finally, the current fix as part of the PR partially resolves the original issue, but allows ```2``` to potentially continue with the added bypass for ```newlineLimit``` that's placed in currently.

````javascript
if (typeof str !== 'string' || str.length <= charLimit && !newlineLimit) {
  return str;
}
````

Since have a little bit of time, I'll rewrite the function instead of performing the quick patch.

File: src/filters/truncate.js [hunk 2]
```diff
@@ -2,30 +2,33 @@
...
-      if (useWordBoundary !== false) {
-        // Find the last word break, but don't look more than 10 characters back.
-        // Make sure we show at least the first 5 characters.
-        var startIndex = Math.max(4, charLimit - 10);
-        var lastSpace = truncated.lastIndexOf(/\s/, startIndex);
-        if (lastSpace !== -1) {
-          truncated = truncated.substring(0, lastSpace);
+      if (truncated.length > limiter) {
+        if (useWordBoundary) {
+          var tmpLimiter = truncated.lastIndexOf(' ', limiter);
```
spadgett: This changes it from looking for any whitespace (tabs, newlines, etc.) to looking specifically for `' '`

src/filters/truncate.js
```diff
@@ -2,30 +2,33 @@
...
       var truncated = str;
+      var limiter = charLimit;
+      var removedCharsLength = charLimit - (removeExtraChars || 0);
 
-      if (charLimit) {
-        truncated = truncated.substring(0, charLimit);
-      }
-
-      if (newlineLimit) {
+      if (newlineLimit && truncated.length > limiter) {
```
spadgett: The intent was to make newlineLimit a hard limit, even if the string is < char limit. `foo\nbar` should get truncated still if I say newline limit 1, char limit 80, even though it's less than 80 characters.

src/filters/truncate.js
```diff
@@ -2,30 +2,33 @@
...
+
+          if (tmpLimiter > 0) {
+            limiter = tmpLimiter;
+          }
         }
+        truncated = truncated.substring(0, limiter);
+      }
+
+      if (removeExtraChars && truncated.length >= removedCharsLength) {
+        truncated = truncate(truncated, removedCharsLength, useWordBoundary);
```
spadgett: If I'm reading this right, won't this remove the entire string if I pass in a long unbroken identifier? Or am I misunderstanding?

Before it would never look more than 10 characters back for a word boundary.

src/filters/truncate.js
```diff
@@ -2,30 +2,33 @@
 
 angular.module('openshiftCommonUI')
   .filter('truncate', function() {
-    return function(str, charLimit, useWordBoundary, newlineLimit) {
-      if (!str) {
+    return function truncate(str, charLimit, useWordBoundary, newlineLimit, removeExtraChars) {
```
spadgett: It's not clear to me from the name with `removeExtraChars` is supposed to do.

src/filters/truncate.js
cdcabrera: Gotcha

spadgett:
@cdcabrera Can you give some examples of what is happening now vs before? It's not clear to me what we're trying to change.

src/filters/truncate.js
cdcabrera: it does, I don't think the original was actually firing with the regex that was in there...  ```/\s/```

cdcabrera:
Can do on some examples... unit tests will also help, I'll start with the tests and work my way over to any remaining examples. As far as the issue, #98 , the fix is based off of I can include a couple of gif animations of related copy

src/filters/truncate.js
cdcabrera: Yeah that one is a doozy, not sure what a good variable name is... could actually spell out ```characters``` but thinking that isn't going to help much.  Could aim for something like ```makeSpaceForTheShowHideLinkPlacement``` =)... Up for recommendations on it

spadgett: Oh I see what you're using it for... Currently the show link is usually in the upper right, so if the line wraps, I'm not sure we've made space. We might have changed that for toasts, though.

src/filters/truncate.js
spadgett: Yeah it looks like indexOf or lastIndexOf doesn't take a regex :(

cdcabrera:
@spadgett we discovered an even more simple fix than what we had worked on originally... changing the ```lastIndexOf``` to use a string fixes the original.

```javascript
var startIndex = Math.max(4, charLimit - 10);
var lastSpace = truncated.lastIndexOf(' ', startIndex);
if (lastSpace !== -1) {
  truncated = truncated.substring(0, lastSpace);
}
```
## Results based on the string indicated in issue #98 

Original Results with ```truncated.lastIndexOf(/\s/, startIndex);```
```
Operation cannot be fulfilled on namespaces
"ups-broker": The system is ensuring all content is removed
from this namespace. Upon completion, this
namespace will automatically be purged by the system.
```

New'ish Results with ```truncated.lastIndexOf(' ', startIndex);```
```
Operation cannot be fulfilled on namespaces
"ups-broker": The system is ensuring all content is removed
from this namespace. Upon completion, this
namespace will automatically be purged by
```

We can still take care of a unit test for it, but a one'ish-character fix seems like it fits more in line with avoiding consumer repo breaks/issues

cdcabrera:
@spadgett caught up with @jwforres 

Decided to go with a bit more of a direct approach by correcting the regex inside ```lastIndexOf```. Doing this causes the word boundary truncate to function appropriately, however it alters the truncate behavior, behavior we may have written around. 

Obviously by now, going with the recommendation of holding off on including this for 3.6

The included updates within this PR currently only include
- an additional check against the parameters
- and the string fix from regex to lastIndexOf

src/filters/truncate.js
```diff
@@ -3,7 +3,7 @@
 angular.module('openshiftCommonUI')
   .filter('truncate', function() {
     return function(str, charLimit, useWordBoundary, newlineLimit) {
-      if (!str) {
+      if (typeof str !== 'string' || typeof charLimit !== 'number') {
```
spadgett: This will ignore `newlineLimit` if `charLimit` isn't set. Not sure if we ever actually do that, but it might be useful with a one line limit and `text-overflow: ellipsis;` where we use CSS to truncate a long, single line.

src/filters/truncate.js
```diff
@@ -22,7 +22,7 @@ angular.module('openshiftCommonUI')
         // Find the last word break, but don't look more than 10 characters back.
         // Make sure we show at least the first 5 characters.
         var startIndex = Math.max(4, charLimit - 10);
-        var lastSpace = truncated.lastIndexOf(/\s/, startIndex);
+        var lastSpace = truncated.lastIndexOf(' ', startIndex);
```
spadgett: It looks to me like this logic will run for even for strings smaller than `charLimit`, or am I wrong? We should check that small strings aren't truncated. That might have been masked before by the incorrect use of a regex for `lastIndexOf`.

spadgett:
> Decided to go with a bit more of a direct approach by correcting the regex inside lastIndexOf. 

👍  sounds good to me.

src/filters/truncate.js
cdcabrera: We can wrap an additional check around it

src/filters/truncate.js
cdcabrera: ah didn't account for an either both, or just one, scenario on ```charLimit``` and ```newlineLimit```. updating the condition

src/filters/truncate.js
cdcabrera: @spadgett we've got a couple of options for handling this. We could completely ignore the word boundary flag if the limit is already met, like so...

```javascript
if (useWordBoundary !== false && truncated.length > charLimit) {
```
Or we could still try to include the word boundary then apply a second condition for removing the 10 characters if the length still isn't below/equal to the ```charLimit```. You have preference?

cdcabrera:
@spadgett unit test updated, behavior documented

When you review the spec file you'll see the difference in behavior when word boundary is activated, [around line 49](https://github.com/openshift/origin-web-common/pull/107/files#diff-e34a231b3d750810423d7140c650a33dR49)

src/filters/truncate.js
```diff
@@ -21,8 +21,12 @@ angular.module('openshiftCommonUI')
       if (useWordBoundary !== false) {
         // Find the last word break, but don't look more than 10 characters back.
         // Make sure we show at least the first 5 characters.
-        var startIndex = Math.max(4, charLimit - 10);
-        var lastSpace = truncated.lastIndexOf(/\s/, startIndex);
+        if (truncated.length > charLimit) {
```
spadgett: Won't this always evaluate to false since we call substring already above?

cdcabrera: Indeed it doesn't fire in most cases, correct... looks like it needs to be corrected because the only way it returns a value is when charLimit is ```null``` which is kind of useless, but the intent appears to be related to how the "hard truncate" is separate from "new line", and "word boundary". Regardless this does help highlight the use of the "-10" characters.

On an earlier version of this we briefly had the "-10" as a parameter with a default of "-10" we could go back to that, or revert it entirely to maintain part of the original behavior?

cdcabrera: Related to the request for a word break going 10 characters back... there is a partially related unit test...

```javascript
line:49     it('should return a string that ends on a word boundary', function() {
```
going to look at adding a related ```expect``` or new test

cdcabrera:
@spadgett 
- unit test updated to be more specific for number of new lines. 
- It appears, yes, because of the "sentence" being used in the unit test, the [word boundary check that's in there appears to accomplish your request](https://github.com/cdcabrera/origin-web-common/blob/f353fa36680ecc234f14dcb4cbd543c62756adbe/test/spec/filters/truncateSpec.js#L58). If you can think of an additional way to handle it I'll update accordingly.
- Added a test that combines both word boundary and new line checks

test/spec/filters/truncateSpec.js
```diff
@@ -0,0 +1,97 @@
...
+
+  it('should return a string that ends on a word boundary', function() {
+
+    var str = $scope.params.str;
+    var charLimit = $scope.params.charLimit;
+    var useWordBoundary = true;
+    var newlineLimit = $scope.params.newlineLimit;
+    var filteredString = $filter('truncate')(str, charLimit, useWordBoundary, newlineLimit);
+
+    expect(filteredString.length).toEqual(188);
```
spadgett: Hm, this looks like a bug? It should not look more than 10 characters back for word boundaries, so I'd expect the smallest length to be 190.

cdcabrera: Can try fiddling with the look back logic on it

cdcabrera: @spadgett 
The alteration around aiming for a specific word boundary within the last 10 characters of a string means the word boundary won't fire in certain cases where a word might be rather lengthy. If we're good with that concept I can implement this logic within the filter that gives a result of ```192``` characters

```javascript
if (useWordBoundary !== false) {

    var lastSpace = truncated.lastIndexOf(' ');
    var startIndex = Math.max(4, charLimit - 10);

    if (lastSpace > startIndex && lastSpace !== -1) {
      truncated = truncated.substring(0, lastSpace);
    }
}
```

src/filters/truncate.js
```diff
@@ -21,8 +21,8 @@ angular.module('openshiftCommonUI')
       if (useWordBoundary !== false) {
 
         var startIndex = Math.max(4, charLimit - 10);
-        var lastSpace = truncated.lastIndexOf(' ', startIndex);
-        if (lastSpace !== -1) {
+        var lastSpace = truncated.lastIndexOf(' ');
+        if (lastSpace > startIndex && lastSpace !== -1) {
```
spadgett: Ah ha, thank you for fixing this :+1:

test/spec/filters/truncateSpec.js
spadgett: Probably should be `>=`. Thank you for catching this. We were using `lastIndexOf` completely incorrectly.

PR #108, opened 2017-06-14
Don't show both a toast and in-page alert for create project

spadgett:
@jwforres I want to make this change, but I'm worried it will break the console integration tests (on DCUT day no less). It looks like we'll show both an in-page alert and toast on create project errors, however.

@jeff-phillips-18 FYI


spadgett:
Looks like this is OK since https://github.com/openshift/origin-web-console/pull/1781 passed Jenkins

PR #112, opened 2017-06-17
Authorize websocket requests with protocol header

liggitt:
Requires https://github.com/openshift/origin/pull/14716

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -612,7 +612,7 @@ angular.module('openshiftCommonServices')
   	}
   };
 
-  this.$get = function($q, $injector, $log, $rootScope, Logger) {
+  this.$get = function($q, $injector, $log, $rootScope, Logger, base64) {
     var authLogger = Logger.get("auth");
     authLogger.log('AuthServiceProvider.$get', arguments);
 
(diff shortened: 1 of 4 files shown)


dist/origin-web-common-services.js
```diff
@@ -693,8 +693,23 @@ angular.module('openshiftCommonServices')
...
+          config.protocols = config.protocols || [];
+
+          // Ensure protocols is an array
+          if (!_.isArray(config.protocols)) {
+            config.protocols = [config.protocols];
+          }
+
+          // Ensure protocols has at least one item in it (for the server to echo back once the bearer protocol is stripped out)
+          if (config.protocols.length == 0) {
+            config.protocols.unshift("undefined");
```
jwforres: could we do something like console.openshift.io.placeholder or something?  I worry that seeing "undefined" in a request handshake might cause someone to open a bug/ticket

jwforres:
Marked as WIP since this depends on another PR that isn't merged

dist/origin-web-common-services.js
liggitt: firefox sends "undefined" today, so I know it works. I guess I don't mind sending a specific placeholder protocol, though it's hard to pick one that I could see upstreaming to kube (which I want to)

liggitt:
@jwforres prereq and upstream are merged, this is ready

liggitt:
Do I need to cut/bump this in the web console repo?

jwforres:
Yeah someone will need to do that

jeff-phillips-18:
There is another change ( #119 ) that requires an update in both origin-web-catalog and origin-web-console before we add to the web console repo.

dist/origin-web-common-services.js
xingxingxia: I'm one of the "worrying" :)
I'm intending to leave comment in card to @liggitt whether "undefined" is indicating something exceptional, and finding you already discussed it. Solved my worrying :) thx

PR #117, opened 2017-06-23
Improve toast notification performance

spadgett:
@jeff-phillips-18 @jwforres Opinion on this change?

I'm a little concerned that we're using `$scope.$watch` on the entire array from `NotificationsService` using a deep comparison. The `$watch` checks are evaluated frequently, especially with our use of websockets, and the notifications array is never trimmed.

### DIFF
--- a/src/components/toast-notifications/toastNotifications.js
+++ b/src/components/toast-notifications/toastNotifications.js
@@ -1,49 +1,79 @@
 'use strict';
 
 angular.module('openshiftCommonUI')
-  .directive('toastNotifications', function(NotificationsService, $timeout) {
+  .directive('toastNotifications', function(NotificationsService, $rootScope, $timeout) {
     return {
       restrict: 'E',
       scope: {},
       templateUrl: 'src/components/toast-notifications/toast-notifications.html',
       link: function($scope) {
-        $scope.notifications = NotificationsService.getNotifications();
+        $scope.notifications = [];
+
+        // A notification is removed if it has hidden set and the user isn't
+        // currently hovering over it.
+        var isRemoved = function(notification) {
+          return notification.hidden && !notification.isHover;
+        };
+
+        var removeNotification = function(notification) {
+          notification.isHover = false;
+          notification.hidden = true;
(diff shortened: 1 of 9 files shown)


src/components/toast-notifications/toastNotifications.js
```diff
@@ -1,49 +1,75 @@
...
+        // Remove items that are now hidden to keep the array from growing
+        // indefinitely. We loop over the entire array each digest loop, even
+        // if everything is hidden, and any watch update triggers a new digest
+        // loop. If the array grows large, it can hurt performance.
+        var removeHiddenToasts = function() {
+          $scope.notifications = _.reject($scope.notifications, isRemoved);
+        };
+
+        var closeNotification = function(notification) {
+          notification.isHover = false;
```
jwforres: is there any chance of setHover getting fired and re-setting this flag while the animation is running?

src/ui-services/notificationsService.js
```diff
@@ -23,6 +23,7 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
       }
 
       notifications.push(notification);
+      $rootScope.$emit('showToastNotification', notification);
```
jwforres: instead of showToastNotification i think it would be cleaner to say newNotification or notificationAdded or something, in case something else was going to want trigger off of this later?

spadgett:
Thanks @jwforres, updated

src/ui-services/notificationsService.js
```diff
@@ -75,7 +76,7 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
     };
 
     // Also handle `addNotification` events on $rootScope, which is used by DataService.
-    $rootScope.$on('addNotification', function(event, data) {
+    $rootScope.$on('NotificationService.addNotification', function(event, data) {
```
jwforres: NotificationsService, plural for consistency

spadgett: Yes, thank you

PR #120, opened 2017-06-27
Use .word-break mixin along with max-width to enable word break within toast notifications.

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1770

screen shot 2017-06-27 at 11 46 05 am
screen shot 2017-06-27 at 11 35 28 am

screen shot 2017-06-27 at 11 33 11 am

### DIFF
--- a/src/styles/_notifications.less
+++ b/src/styles/_notifications.less
@@ -8,7 +8,13 @@
 }
 
 .toast-notifications-list-pf {
+  @media(max-width: @screen-xs-max) {
+    // Set max-width when < 769, so that word-break functions on Firefox & IE.
+    // Subtract 40px to align edges with content within container-fluid that has left & right 20px margin
+    max-width: calc(~"100% - 40px");
+  }
   top: 28px;
+  .word-break();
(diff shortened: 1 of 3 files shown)


sg00dwin:
@rhamilto, @jwforres

src/styles/_notifications.less
```diff
@@ -8,7 +8,11 @@
 }
 
 .toast-notifications-list-pf {
+  @media(max-width: @screen-xxs-max) {
+    max-width: calc(100% - - 40px); // Set max-width when < 769, so that word-break functions on Firefox & IE.
```
rhamilto: Double minus?  @screen-xxs-max = 479, not 769.  Add additional comment on why 40px?

sg00dwin: the 40px accounts for the 20px left & right margin from `.container-fluid` 

Found that these are equal, but I'll switch to the other
calc(100% - - 40px) === calc(~"100% - 40px")

meant to use var @screen-xs-max, will update.

sg00dwin:
@rhamilto updated pr

PR #127, opened 2017-07-12
Prevent double scrollbars when lots of options within ui-select on bind form

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1814

m1
m3


sg00dwin:
@spadgett

PR #129, opened 2017-07-24
Add opts.skipDigest for DataService.watch

benjaminapetersen:
- provides mechanism to keep WebSockets from spamming the `$digest` loop
- flips `invokeApply` passed to the internal `$timeout` calls,
  stopping a `$digest` loop & avoiding a render.
- `DataService._watchOpOnMessage` also honors the `suppressDigest` flag,
  which is plumbed down through `.watch` -> `.startListOp` -> `._listOpComplete` -> `._startWatchOp` -> `._watchOpOnMessage`

example usage:

```javascript 
DataService.watch(resource, context, (data) => {
  // optionally call $applyAsync here in controlling code
  // to manually run a $digest
  $scope.$applyAsync(() => {
   // do stuff with data here
  });
}, { suppressDigest: true });

// or debounced 
DataService.watch(resource, context, _.debounce((data) => {
  // $apply here, inside debounce
}, 400), { suppressDigest: true });
```

@spadgett @jwforres for review

### DIFF
--- a/src/services/dataService.js
+++ b/src/services/dataService.js
@@ -602,6 +602,9 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
 //            pollInterval: in milliseconds, how long to wait between polling the server
 //                    only applies if poll=true.  Default is 5000.
 //            http:   similar to .get, etc. at this point, only used to pass http.params for filtering
+//            skipDigest: will skip the $apply & avoid triggering a digest loop
+//                    if set to `true`.  Is intentionally the inverse of the invokeApply
+//                    arg passed to $timeout (due to default values).
 //            errorNotification: will popup an error notification if the API request fails (default true)
 // returns handle to the watch, needed to unwatch e.g.
 //        var handle = DataService.watch(resource,context,callback[,opts])
@@ -609,6 +612,7 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
   DataService.prototype.watch = function(resource, context, callback, opts) {
     resource = APIService.toResourceGroupVersion(resource);
     opts = opts || {};
+    var invokeApply =  !opts.skipDigest;
     var key = this._uniqueKey(resource, null, context, _.get(opts, 'http.params'));
     if (callback) {
       // If we were given a callback, add it
@@ -636,7 +640,7 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
       if (callback) {
         $timeout(function() {
           callback(self._data(key));
-        }, 0);
+        }, 0, invokeApply);
       }
     }
     else {
@@ -648,7 +652,7 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
             if (resourceVersion === self._resourceVersion(key)) {
               callback(self._data(key)); // but just in case, still pull from the current data map
             }
-          }, 0);
+          }, 0, invokeApply);
         }
       }
       if (!this._listInFlight(key)) {
@@ -1111,6 +1115,8 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
 
   DataService.prototype._watchOpOnMessage = function(resource, context, opts, event) {
     var key = this._uniqueKey(resource, null, context, _.get(opts, 'http.params'));
+    opts = opts || {};
+    var invokeApply = !opts.skipDigest;
     try {
       var eventData = $.parseJSON(event.data);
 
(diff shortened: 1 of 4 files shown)


src/services/dataService.js
```diff
@@ -602,13 +602,17 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
 //            pollInterval: in milliseconds, how long to wait between polling the server
 //                    only applies if poll=true.  Default is 5000.
 //            http:   similar to .get, etc. at this point, only used to pass http.params for filtering
+//            suppressDigest: will skip the $apply & avoid triggering a digest loop
```
spadgett: `skipDigest` maybe? Or possibly `skipApply` to be consistent with `$timeout` naming. (No strong preference.)

src/services/dataService.js
```diff
@@ -602,13 +602,17 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
...
+//                    if set to `true`.  Is intentionally the inverse of the invokeApply
+//                    arg passed to $timeout (due to default values).
 //            errorNotification: will popup an error notification if the API request fails (default true)
 // returns handle to the watch, needed to unwatch e.g.
 //        var handle = DataService.watch(resource,context,callback[,opts])
 //        DataService.unwatch(handle)
   DataService.prototype.watch = function(resource, context, callback, opts) {
     resource = APIService.toResourceGroupVersion(resource);
     opts = opts || {};
+    var invokeApply = opts.hasOwnProperty('suppressDigest') && opts.suppressDigest ? false : true;
```
spadgett: I think this could just be

```js
var invokeApply = !_.get(opts, 'suppressDigest');
```

src/services/dataService.js
```diff
@@ -1110,6 +1114,7 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
   };
 
   DataService.prototype._watchOpOnMessage = function(resource, context, opts, event) {
+    var invokeApply = opts.hasOwnProperty('suppressDigest') && opts.suppressDigest ? false : true;
```
spadgett: Same comment as above.

src/services/dataService.js
benjaminapetersen: I'm good for a rename.  I wonder if `manualApply` or `manualDigest` is preferable even, to make controller code easiest to read:

```JavaScript 
DataService.watch(resource, () = {
  // manually do updates
}, { skipApply: true });

DataService.watch(resource, () = {
  // manually do updates
}, { manuallyApply: true });
```
Thoughts?

src/services/dataService.js
benjaminapetersen: Nice thats better.  Will update.

spadgett: Actually could just be

```js
var invokeApply = !opts.suppressDigest;
```

since opts is never null at this point.

src/services/dataService.js
spadgett: To me `skipDigest` is the most obvious / unambiguous. `apply` is such a general word that I worry that it's not clear what it means.

benjaminapetersen: Yeah, digest is a lot more specific. Sounds good to me.

src/services/dataService.js
```diff
@@ -602,13 +602,17 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
...
+//                    if set to `true`.  Is intentionally the inverse of the invokeApply
+//                    arg passed to $timeout (due to default values).
 //            errorNotification: will popup an error notification if the API request fails (default true)
 // returns handle to the watch, needed to unwatch e.g.
 //        var handle = DataService.watch(resource,context,callback[,opts])
 //        DataService.unwatch(handle)
   DataService.prototype.watch = function(resource, context, callback, opts) {
     resource = APIService.toResourceGroupVersion(resource);
     opts = opts || {};
+    var invokeApply =  !_.get(opts, 'skipDigest');
```
spadgett: I realized after making my comment that it could just be `!opts.skipDigest` since `opts` is never null :)

benjaminapetersen: Haha, it can... oi, my first pass was convoluted :smile:  
Was overthinking it, worried I'd accidentally wipe out the default `true` for `$timeout`.

Updated.  This does work, I have tested the following on the pods page:

```javascript 
watches.push(DataService.watch("pods", context, function(pods) {
       // do things that won't update UI
}, { skipDigest: true }));
```
And am manually deploying pods from other pages & can verify no updates.

benjaminapetersen:
Yup, will do

PR #136, opened 2017-08-03
Recently Viewed Projects Service

dtaylor113:
Added RecentlyViewedProjectsService which uses localStorage to persist recently viewed project uids.   Updated `ProjectsService.get(projectName)` to save project's uid in RecentlyViewedProjectsService.  Added  `ProjectsService.create(...)`
Unit tests included.
This is the first PR towards resolving issue [console:#1895](https://github.com/openshift/origin-web-console/issues/1895)

### DIFF
--- a/src/components/create-project/createProject.js
+++ b/src/components/create-project/createProject.js
@@ -11,7 +11,7 @@ angular.module("openshiftCommonUI")
         isDialog: '@'
       },
       templateUrl: 'src/components/create-project/createProject.html',
-      controller: function($scope, $filter, $location, DataService, NotificationsService, displayNameFilter) {
+      controller: function($scope, $location, ProjectsService, NotificationsService, displayNameFilter) {
         if(!($scope.submitButtonLabel)) {
           $scope.submitButtonLabel = 'Create';
         }
@@ -25,27 +25,18 @@ angular.module("openshiftCommonUI")
         $scope.createProject = function() {
           $scope.disableInputs = true;
           if ($scope.createProjectForm.$valid) {
-            DataService
-              .create('projectrequests', null, {
-                apiVersion: "v1",
-                kind: "ProjectRequest",
-                metadata: {
-                  name: $scope.name
-                },
-                displayName: $scope.displayName,
-                description: $scope.description
-              }, $scope)
-              .then(function(data) {
+            ProjectsService.create($scope.name, $scope.displayName, $scope.description)
+              .then(function(project) {
                 // angular is actually wrapping the redirect action
                 var cb = $scope.redirectAction();
                 if (cb) {
-                  cb(encodeURIComponent(data.metadata.name));
+                  cb(encodeURIComponent(project.metadata.name));
                 } else {
-                  $location.path("project/" + encodeURIComponent(data.metadata.name) + "/create");
+                  $location.path("project/" + encodeURIComponent(project.metadata.name) + "/create");
                 }
                 NotificationsService.addNotification({
                   type: "success",
-                  message: "Project \'"  + displayNameFilter(data) + "\' was successfully created."
+                  message: "Project \'"  + displayNameFilter(project) + "\' was successfully created."
                 });
               }, function(result) {
                 $scope.disableInputs = false;
@@ -76,6 +67,6 @@ angular.module("openshiftCommonUI")
         };
 
         $scope.$on("$destroy", hideErrorNotifications);
-      },
+      }
     };
   });
--- a/src/services/projects.js
+++ b/src/services/projects.js
@@ -2,7 +2,7 @@
 
 angular.module('openshiftCommonServices')
   .factory('ProjectsService',
-    function($location, $q, AuthService, DataService, annotationNameFilter, AuthorizationService) {
+    function($location, $q, AuthService, DataService, annotationNameFilter, AuthorizationService, RecentlyViewedProjectsService) {
 
 
       var cleanEditableAnnotations = function(resource) {
@@ -37,6 +37,7 @@ angular.module('openshiftCommonServices')
                                         .then(function() {
                                           context.project = project;
                                           context.projectPromise.resolve(project);
+                                          RecentlyViewedProjectsService.addProjectUID(project.metadata.uid);
                                           // TODO: fix need to return context & projectPromise
                                           return [project, context];
                                         });
@@ -66,6 +67,23 @@ angular.module('openshiftCommonServices')
           update: function(projectName, data) {
             return DataService
                     .update('projects', projectName, cleanEditableAnnotations(data), {projectName: projectName}, {errorNotification: false});
+          },
+          create: function(name, displayName, description) {
+            var projectRequest = {
+              apiVersion: "v1",
+              kind: "ProjectRequest",
+              metadata: {
+                name: name
+              },
+              displayName: displayName,
+              description: description
+            };
+            return DataService
+              .create('projectrequests', null, projectRequest, {})
+              .then(function(project) {
+                RecentlyViewedProjectsService.addProjectUID(project.metadata.uid);
+                return project;
+              });
           },
           canCreate: function() {
             return DataService.get("projectrequests", null, {}, { errorNotification: false});
--- a/src/services/recentlyViewedProjects.js
+++ b/src/services/recentlyViewedProjects.js
@@ -1,0 +1,63 @@
+'use strict';
+
+angular.module("openshiftCommonServices")
+  .service("RecentlyViewedProjectsService", function($filter){
+
+    var recentlyViewedProjsKey = "openshift/recently-viewed-project-uids";
+
+    var getProjectUIDs = function() {
+      var recentlyViewed = localStorage.getItem(recentlyViewedProjsKey);
+      return recentlyViewed ? JSON.parse(recentlyViewed) : [];
+    };
+
+    var addProjectUID = function(uid) {
+      var recentlyViewed = getProjectUIDs();
+
+      // add to front of list
+      recentlyViewed.unshift(uid);
+
+      // no dups
+      recentlyViewed = _.uniq(recentlyViewed);
+
+      // limit to 5 items
+      recentlyViewed = _.take(recentlyViewed, 5);
+
+      setRecentlyViewedProjects(recentlyViewed);
+    };
+
+    var clear = function() {
+      localStorage.removeItem(recentlyViewedProjsKey);
+    };
+
+    var setRecentlyViewedProjects = function(recentlyViewed) {
+      localStorage.setItem(recentlyViewedProjsKey, JSON.stringify(recentlyViewed));
+    };
+
+    var orderByMostRecentlyViewed = function(projects) {
+      var recentlyViewedProjects = [];
+      var recentlyViewedIds = getProjectUIDs();
+
+      // remove mostRecentlyViewed projects
+      _.each(recentlyViewedIds, function(uid) {
+        var proj = _.remove(projects, function(project) {
+          return project.metadata.uid === uid;
+        })[0];
+        if(proj !== undefined) {
+          recentlyViewedProjects.push(proj);
+        }
+      });
+
+      // order by creationDate
+      projects =  $filter('orderObjectsByDate')(projects, true);
+
+      // return array where moveRecentlyViewed is first, then sorted by creationDate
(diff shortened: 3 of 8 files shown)


spadgett:
@dtaylor113 Ready for review?

dtaylor113:
@spadgett , yes!

jeff-phillips-18:
wondering if only counting calls to ProjectService.get is the right thing to do. Any calls to DataService.get("projects", projectName... will not be counted. What about creating a project? Does that end up doing a get or should it be added on the creation call?

spadgett:
+1 to create project. I was thinking about that as well. I'd like to add a `ProjectsService.create()`. We are creating projects in many places now and have copied code for it (e.g. https://github.com/openshift/origin-web-console/pull/1893). Then we update recent projects there, too.

FWIW, every page in the console inside a project (that I know of) uses `ProjectsService.get()`, so I think it's a reasonable start. @jeff-phillips-18 do you have ideas where else we could do it?

dtaylor113:
> I'd like to add a ProjectsService.create()

Hi @spadgett, ok if this is done in a follow up PR?

I was thinking of adding the `RecentlyViewedProjectsService.addProject(project.metadata.uid)` to dataService.create() where resource is 'projectrequests' : https://github.com/openshift/origin-web-common/blob/master/src/services/dataService.js#L308

jeff-phillips-18:
What about when adding a catalog item to a project? Should that then bump the project to the top of the list? If it is a new project (creating during the add to project), looks like it is using DataService.create("projectrequests"...

dtaylor113:
> Any calls to DataService.get("projects", projectName... will not be counted

Hi @jeff-phillips-18, the recently viewed projects should only be counted when one views a single project or creates a project.  I did miss the create usecase.   Viewing a single project does result in a ProjectsService.get().  I do not think `DataService.get("projects", projectName` should trigger the recently viewed.

dtaylor113:
[This](https://github.com/openshift/origin-web-common/pull/136/files#diff-bca150427bbb2ec6465ac715cfc1ea05R322) fixed the create usecase.

spadgett:
I think it's OK in a follow up.

dtaylor113:
> What about when adding a catalog item to a project? Should that then bump the project to the top of the list?

Good question, it's not really recently viewed, more like recently used, but that's just semantics.  That is easy enough to add.  @spadgett whatdoyathink?

> If it is a new project (creating during the add to project), looks like it is using DataService.create("projectrequests"...

Yes, with the fix I just added, it does go to the top of the projects list.

spadgett:
I'm a little hesitant to start putting recently viewed logic into `DataService`. Would rather try to keep it together in `ProjectsService` if possible (and maybe just a couple other places like provision service).

dtaylor113:
DataService is the one place all project creations go through.   I can move it to common/create-project and catalog/order-service & create-from-builder wizards when creating a new project.   I'm not sure that covers all the 'create project' usescases though.

spadgett:
We're trying to keep `DataService` as a generic interface for interacting with the API that other clients can use. I'm really hesitant to start adding side effects to `DataService` calls.

dtaylor113:
> We're trying to keep DataService as a generic interface for interacting with the API that other clients can use. I'm really hesitant to start adding side effects to DataService calls.

Ok, np, I've already moved it to the other areas I mentioned.

dtaylor113:
> What about when adding a catalog item to a project? Should that then bump the project to the top of the list?

@spadgett do you think we should do this?

dtaylor113:
Ok, moved counting 'create project' from DataService to 'create-project' component.  Creating and counting a project for most recently viewed (mrv) for service items and builder images handled in origin-catalog.  Creating a project for a Template will be handled in origin-console.

dtaylor113:
@spadgett please don't merge yet, working on the ProjectServices.create() since I'm modifying all the endpoints already.

src/services/recentlyViewedProjects.js
```diff
@@ -0,0 +1,60 @@
...
+      _.forEach(recentlyViewedIds, function(uid) {
+        var proj = _.remove(projects, function(project) {
+          return project.metadata.uid === uid;
+        })[0];
+        if(proj !== undefined) {
+          recentlyViewedProjects.push(proj);
+        }
+      });
+
+      return recentlyViewedProjects.concat(projects);
```
spadgett: We should fallback to either most recently-created or alphabetical for the remaining projects.

src/services/recentlyViewedProjects.js
```diff
@@ -0,0 +1,60 @@
...
+
+    var setRecentlyViewedProjects = function(recentlyViewed) {
+      localStorage.setItem(recentlyViewedProjsKey, JSON.stringify(recentlyViewed));
+    };
+
+    var orderByMostRecentlyViewed = function(projects) {
+      var recentlyViewedProjects = [];
+      var recentlyViewedIds = getProjects();
+
+      _.forEach(recentlyViewedIds, function(uid) {
```
spadgett: nit: `_.each` to be consistent with other code

src/services/recentlyViewedProjects.js
dtaylor113: Though we agreed that it would be ordered by 'most recently viewed' and then by 'most recently created'.  Are we changing our minds on this?

bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "origin-web-common",
-  "version": "0.0.43",
+  "version": "0.0.44",
```
dtaylor113: Bumped to 45.  Going forward I will do separate PRs for version bumps.

src/services/projects.js
```diff
@@ -67,6 +68,18 @@ angular.module('openshiftCommonServices')
...
+            var deferred = $q.defer();
+            DataService
+              .create('projectrequests', null, object, context)
+              .then(function(project) {
+                RecentlyViewedProjectsService.addProject(project.metadata.uid);
+                deferred.resolve(project);
+              }, function(result) {
+                deferred.reject(result);
+              });
+            return deferred.promise;
```
dtaylor113: Good suggestions, fixed :-)

src/services/recentlyViewedProjects.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+angular.module("openshiftCommonServices")
+  .service("RecentlyViewedProjectsService", function(){
+
+    var recentlyViewedProjsKey = "common-recently-viewed-projects";
```
dtaylor113: Fixed

src/services/recentlyViewedProjects.js
```diff
@@ -0,0 +1,60 @@
...
+
+    var getProjects = function() {
```
dtaylor113: Fixed

src/services/recentlyViewedProjects.js
```diff
@@ -0,0 +1,60 @@
...
+      var recentlyViewed = localStorage.getItem(recentlyViewedProjsKey);
+      return recentlyViewed ? JSON.parse(recentlyViewed) : [];
+    };
+
+    var addProject = function(uid) {
```
dtaylor113: Fixed

src/services/recentlyViewedProjects.js
```diff
@@ -0,0 +1,60 @@
...
+      var recentlyViewed = getProjects();
+
+      // if previously viewed, remove from list
+      _.remove(recentlyViewed, function(rvUID) {
+        return rvUID === uid;
+      });
```
dtaylor113: Fixed

src/services/recentlyViewedProjects.js
dtaylor113: Fixed

src/services/recentlyViewedProjects.js
spadgett: > Are we changing our minds on this?

No.

So it's sorted first by the caller? I'd rather pass the unsorted list and do all of the sorting here.

dtaylor113: Project Summary in Catalog does this:
```
    var projects: any = _.toArray(projectData.by('metadata.creationTimestamp'));
    var orderByDate: any = this.$filter('orderObjectsByDate');
    this.ctrl.projects = orderByDate(projects, true);
```
So you are saying you'd like to move the sorting by creationTimestamp into the recentlyViewedProjects.orderByMostRecentlyViewed()?

spadgett: > So you are saying you'd like to move the sorting by creationTimestamp into the recentlyViewedProjects.orderByMostRecentlyViewed()?

Yes, as the fallback sort to the remaining projects.

bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "origin-web-common",
-  "version": "0.0.44",
+  "version": "0.0.45",
```
spadgett: Do you mind also removing the version bump from this PR per @jeff-phillips-18's comments in the catalog PR?

src/services/recentlyViewedProjects.js
dtaylor113: Ok, fixed

bower.json
dtaylor113: Removed the version bump.

src/services/recentlyViewedProjects.js
```diff
@@ -0,0 +1,63 @@
...
+    var setRecentlyViewedProjects = function(recentlyViewed) {
+      localStorage.setItem(recentlyViewedProjsKey, JSON.stringify(recentlyViewed));
+    };
+
+    var orderByMostRecentlyViewed = function(projects) {
+      var recentlyViewedProjects = [];
+      var recentlyViewedIds = getProjectUIDs();
+
+      // first order by creationDate
+      projects =  $filter('orderObjectsByDate')(projects, true);
```
spadgett: Suggest we move this below the `_.each` so that we sort the smaller array after the remove the other recently-viewed projects.

dtaylor113: Ok, fixed

src/components/create-project/createProject.js
```diff
@@ -11,7 +11,8 @@ angular.module("openshiftCommonUI")
         isDialog: '@'
       },
       templateUrl: 'src/components/create-project/createProject.html',
-      controller: function($scope, $filter, $location, DataService, NotificationsService, displayNameFilter) {
+      controller: function($scope, $filter, $location, DataService, ProjectsService, NotificationsService,
+                           RecentlyViewedProjectsService, displayNameFilter) {
```
jeff-phillips-18: It appears DataService is no longer used

dtaylor113: Removed DataService, $filter, and RecentlyViewedProjectsService.

dtaylor113:
> What about when adding a catalog item to a project? Should that then bump the project to the top of the list?

Good question, it's not really recently viewed, more like recently used, but that's just semantics.  That is easy enough to add.  @spadgett whatdoyathink?

spadgett:
> Good question, it's not really recently viewed, more like recently used, but that's just semantics. That is easy enough to add. @spadgett whatdoyathink?

I think it's reasonable, although would it cause the project list in the sidebar to shuffle ever? I imagine the recently viewed items are reevaluated on project watch updates?

spadgett:
I think it'd be odd if the projects ever reordered seemingly randomly while I was still on the same page.

dtaylor113:
> I think it's reasonable, although would it cause the project list in the sidebar to shuffle ever? I imagine the recently viewed items are reevaluated on project watch updates?

Yes, recently viewed are reevaluated on project watch updates, however we only add a project to 'recently viewed' on get and create, we didn't talk about update (ProjectsServices.update(..))

spadgett:
Right, but if we change it to update recently viewed on provision, some random time later the projects on the right will shuffle

dtaylor113:
> Right, but if we change it to update recently viewed on provision, some random time later the projects on the right will shuffle

Oh, ok, 'cause provisioning may take some time to complete.

spadgett:
The provision doesn't trigger a watch update of the project list. Only a new project or an update to an existing project will trigger the update, at which time they'll get reordered. This might happen at any time later (minutes or longer or maybe not at all).

We could add an event from the RecentlyViewedProjectService when it changes so we can reorder immediately. I can't decide if it would still be weird to see the items on the right shuffle, but at least it would happen right away on provision.

PR #145, opened 2017-08-09
Support PartialObjectMetadataList in DataService.list()

spadgett:
Let clients optionally request only object metadata when calling DataService.list().

See https://bugzilla.redhat.com/show_bug.cgi?id=1471033

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -1210,6 +1210,15 @@ angular.module('openshiftCommonServices')
 angular.module('openshiftCommonServices')
 .factory('DataService', function($cacheFactory, $http, $ws, $rootScope, $q, API_CFG, APIService, Logger, $timeout, base64, base64util) {
 
+  // Accept PartialObjectMetadataList. Unfortunately we can't use the Accept
+  // header to fallback to JSON due to an API server content negotiation bug.
+  // https://github.com/kubernetes/kubernetes/issues/50519
+  //
+  // This is a potential version skew issue for when the web console runs in
+  // a pod where we potentially need to support different server versions.
+  // https://trello.com/c/9oaUh8xP
+  var ACCEPT_PARTIAL_OBJECT_METADATA_LIST = 'application/json;as=PartialObjectMetadataList;v=v1alpha1;g=meta.k8s.io';
+
   function Data(array) {
     this._data = {};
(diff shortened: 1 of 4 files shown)


spadgett:
Travis failure is unrelated, fixed by #146

spadgett:
Thanks @jeff-phillips-18 for tracking down the test failure

jeff-phillips-18:
Build is passing now

spadgett:
I'd added `application/json` as a fallback to the `Accept` header in case the server doesn't support partial object metadata

spadgett:
Unfortunately the JSON fallback doesn't work, at least not against 3.6. I'm going to do some more testing with different Accept headers in 3.7.

@liggitt You'll be shocked to hear the Accept header handling in the API server isn't 100% to the HTTP 1.1 spec.

```json
{
  "metadata": {},
  "status": "Failure",
  "message": "only the following media types are accepted: application/json, application/yaml, application/vnd.kubernetes.protobuf",
  "reason": "NotAcceptable",
  "code": 406
}
```

screen shot 2017-08-11 at 7 58 23 am

spadgett:
@liggitt 

This works fine:

```
Accept: application/BOGUS, application/json
```

But this causes an error:

```
Accept: application/json;as=BOGUS;v=v1alpha1;g=meta.k8s.io, application/json
```

Swapping the order like below gives me back the full JSON (which I believe is wrong per the spec).

```
Accept: application/json, application/json;as=PartialObjectMetadataList;v=v1alpha1;g=meta.k8s.io
```

https://tools.ietf.org/html/rfc7231#section-5.3.2

```
   Media ranges can be overridden by more specific media ranges or
   specific media types.  If more than one media range applies to a
   given type, the most specific reference has precedence.  For example,

     Accept: text/*, text/plain, text/plain;format=flowed, */*

   have the following precedence:

   1.  text/plain;format=flowed

   2.  text/plain

   3.  text/*

   4.  */*
```

liggitt:
Mind opening an upstream issue for that?

spadgett:
@liggitt https://github.com/kubernetes/kubernetes/issues/50519

spadgett:
I've removed the JSON fallback since it's not working. @jeff-phillips-18 One more look? Should be the same as the original PR with some additional comments.

PR #150, opened 2017-08-21
Cache project list

spadgett:
Cache the list of projects to avoid requesting them again and again on every page transition.

See https://github.com/openshift/origin-web-console/issues/1923

### DIFF
--- a/src/services/projects.js
+++ b/src/services/projects.js
@@ -2,8 +2,28 @@
 
 angular.module('openshiftCommonServices')
   .factory('ProjectsService',
-    function($location, $q, AuthService, DataService, annotationNameFilter, AuthorizationService, RecentlyViewedProjectsService) {
+           function($location,
+                    $q,
+                    $rootScope,
+                    AuthService,
+                    AuthorizationService,
+                    DataService,
+                    Logger,
+                    RecentlyViewedProjectsService,
+                    annotationNameFilter) {
 
+      // Cache project data when we can so we don't request it on every page load.
+      var cachedProjectData;
+      var cachedProjectDataIncomplete = false;
+
+      var clearCachedProjectData = function() {
+        Logger.debug('ProjectsService: clearing project cache');
+        cachedProjectData = null;
+        cachedProjectDataIncomplete = false;
+      };
+
+      AuthService.onUserChanged(clearCachedProjectData);
+      AuthService.onLogout(clearCachedProjectData);
 
       var cleanEditableAnnotations = function(resource) {
         var paths = [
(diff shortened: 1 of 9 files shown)


spadgett:
@jeff-phillips-18 Still WIP, but would be good to get your input. I'm working on the corresponding changes in origin-web-catalog and origin-web-console.

src/services/projects.js
```diff
@@ -64,10 +86,42 @@ angular.module('openshiftCommonServices')
...
+          list: function(forceRefresh) {
+            if (cachedProjectData && !forceRefresh) {
+              Logger.debug("ProjectsService: returning cached project data");
+              return $q.when(cachedProjectData);
+            }
+
+            Logger.debug("ProjectsService: listing projects, force refresh", forceRefresh);
+            return DataService.list('projects', {}).then(function(projectData) {
+              cachedProjectData = projectData;
+              return projectData;
```
jeff-phillips-18: Should this watch for updates to the projects so that future calls get any updates that have occurred since the last retrieval?

src/services/projects.js
```diff
@@ -64,10 +86,42 @@ angular.module('openshiftCommonServices')
...
+            });
+          },
+
+          watch: function(context, callback) {
+            // Wrap `DataService.watch` so we can update the cached projects
+            // list on changes. TODO: We might want to disable watches entirely
+            // if we know the project list is large.
+            return DataService.watch("projects", context, function(projectData) {
+              cachedProjectData = projectData;
+              callback(projectData);
```
jeff-phillips-18: A call to watch could return the current cachedProjectData. And invoke the callback again when projects change. This would be a bit more complex as this would need to handle unwatch as well.

src/services/projects.js
spadgett: I think a global watch on projects will make the scale problem worse :(

src/services/projects.js
spadgett: Sounds good, I can make that change

src/services/projects.js
jeff-phillips-18: OK, so pages that expect to have the latest data would need to use the force flag then. Is there a plan to revisit the pages that are using this function to determine if the projects data could have changed and been the reason for the navigation? For instance, I add to a project and that takes me to another page. That page would expect a list call to return the latest data but it may not if no watcher was set and the force flag was not set.

spadgett: I think it's reasonable for changes outside of the web console to make the user refresh to get the latest list. We could instead re-list if the projects are fewer than a certain cutoff. But for extremely large lists or lists that fail, I don't see a good alternative. Trying to relist every page causes the console to become unusable.

For changes that happen in the web console, we can update the cached list directly. Let me make sure I'm doing that everywhere we can.

I'm looking at all the places we list or watch projects now, and I'm updating them to use the ProjectsService methods.

In 3.8, we'll be able to limit how many results we get back from the server, but not yet in 3.7.

src/services/projects.js
spadgett: Actually I'm thinking it's better for callers to list first and then watch only if projects are less than some cutoff.

spadgett:
@jeff-phillips-18 Changed to keep the cached list updated on get/create/update/delete in console

PR #151, opened 2017-08-22
Allow OAuth scope(s) to be configured

david-martin:
Add `OAuthScope` getter/setter to the RedirectLoginService.
This change allows specifying the 'scope' parameter when requesting an
authorization code & token.
This is necessary if using an OAuth client that doesn't allow requesting
tokens with the default 'user:full' scope.
e.g. A ServiceAccount OAuth Client https://docs.openshift.com/container-platform/3.6/architecture/additional_concepts/authentication.html#service-accounts-as-oauth-clients

*Example Usage*

```js
angular
  .module('myApp', [
    'openshiftCommonServices'
  ])
  .config(['RedirectLoginServiceProvider', function (RedirectLoginServiceProvider) {
      RedirectLoginServiceProvider.OAuthScope('user:info user:check-access');
  }])
```

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -3081,6 +3081,7 @@ angular.module('openshiftCommonServices')
   var _oauth_authorize_uri = "";
   var _oauth_token_uri = "";
   var _oauth_redirect_uri = "";
+  var _oauth_scope = "";
 
   this.OAuthClientID = function(id) {
     if (id) {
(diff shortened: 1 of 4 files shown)


david-martin:
@spadgett I hit a problem when switching from an `oauthclient` resource in OpenShift to a `serviceaccount` based OAuth Client where only certain scopes could be used when retrieving a token.
More info in the PR description.

Does this seem like a reasonable change?

spadgett:
@liggitt PTAL

liggitt:
LGTM

david-martin:
Thanks @spadgett @liggitt.
Any chance of this being merged and published to bower?

spadgett:
@david-martin master is closed except for blocker bug fixes allowed until Monday. We should be able to merge and cut a release then. (Don't hesitate to ping me if I forget.)

david-martin:
Thanks @spadgett @jeff-phillips-18 @liggitt :+1:

PR #152, opened 2017-08-22
Added secondary sort by displayName to RecentlyViewedProjectsService

dtaylor113:
Added secondary sort for 'displayName' (case insensitive).

### DIFF
--- a/src/services/recentlyViewedProjects.js
+++ b/src/services/recentlyViewedProjects.js
@@ -47,17 +47,23 @@ angular.module("openshiftCommonServices")
         }
       });
 
-      // order by creationDate
-      projects =  $filter('orderObjectsByDate')(projects, true);
+      // second sort by case insensitive displayName
+      projects = _.sortBy(projects, function(project) {
+        return $filter('displayName')(project).toLowerCase();
+      });
 
-      // return array where moveRecentlyViewed is first, then sorted by creationDate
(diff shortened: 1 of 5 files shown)


src/services/recentlyViewedProjects.js
```diff
@@ -47,17 +47,28 @@ angular.module("openshiftCommonServices")
         }
       });
 
-      // order by creationDate
-      projects =  $filter('orderObjectsByDate')(projects, true);
+      if (secondSort === 'displayName') {
```
spadgett: I'd rather just always use display name. Do we have a place where we need creation date? We should be consistent everywhere if possible.

dtaylor113:
> Do we have a place where we need creation date?

Yes, the Project Summary side panel is sorted by creation date.

spadgett:
> Yes, the Project Summary side panel is sorted by creation date.

I don't see a reason to sort them differently. I think it should be most recent, and then fall back to display name everywhere.

src/services/recentlyViewedProjects.js
dtaylor113: Fixed

dtaylor113:
Hi @spadgett, ok, changed it so that secondary sort is always case insensitive displayName

spadgett:
Thanks @dtaylor113 LGTM

PR #155, opened 2017-08-25
Prevent modal popup from removing other resize handlers

spadgett:
@rhamilto This should fix the nav problem you were seeing.

### DIFF
--- a/src/components/origin-modal-popup/origin-modal-popup.js
+++ b/src/components/origin-modal-popup/origin-modal-popup.js
@@ -81,7 +81,10 @@ angular.module("openshiftCommonUI").component("originModalPopup", {
 
     function onHide() {
       hideModalBackdrop();
-      angular.element($window).off('resize', debounceResize);
+      if (debounceResize) {
+        angular.element($window).off('resize', debounceResize);
+        debounceResize = null;
+      }
     }
 
     ctrl.$onChanges = function (changeObj) {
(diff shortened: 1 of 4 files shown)


spadgett:
@jeff-phillips-18 If the handler we pass in as the second arg is null, it removes all resize listeners.

jeff-phillips-18:
I see. Probably at init the debounceResize isn't set yet. The $onChanges should probably be checking for first change or that  newval !== oldval before calling onHide.

spadgett:
We also call onHide onDestroy

jeff-phillips-18:
But the shown value should be correct at that point and that would mean the debounceResize is set. I'm not disagreeing with the fix, just thinking of how this went wrong to begin with ;)

spadgett:
> But the shown value should be correct at that point and that would mean the debounceResize is set. 

It looks like that's not resize listener is not added until the dialog becomes visible (which is good). So if the component is destroyed without showing the dialog, it will remove all listeners.

Either way I think it's good to have the guard so we never accidentally remove other listeners.

src/components/origin-modal-popup/origin-modal-popup.js
```diff
@@ -81,7 +81,10 @@ angular.module("openshiftCommonUI").component("originModalPopup", {
 
     function onHide() {
       hideModalBackdrop();
-      angular.element($window).off('resize', debounceResize);
+      if (debounceResize) {
+        angular.element($window).off('resize', debounceResize);
+        debounceResize = null;
```
spadgett: @jeff-phillips-18 I updated the PR to also null out `debounceResize`

PR #158, opened 2017-08-25
Add annotation for logging visualization url

jcantrill:
Add annotation filter for logging.ui.hostname

### DIFF
--- a/src/filters/annotation.js
+++ b/src/filters/annotation.js
@@ -24,6 +24,7 @@ angular.module('openshiftCommonUI')
       "jenkinsBuildURL":          ["openshift.io/jenkins-build-uri"],
       "jenkinsLogURL":            ["openshift.io/jenkins-log-url"],
       "jenkinsStatus":            ["openshift.io/jenkins-status-json"],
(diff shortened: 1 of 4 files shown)


src/filters/annotation.js
```diff
@@ -24,6 +24,7 @@ angular.module('openshiftCommonUI')
       "jenkinsBuildURL":          ["openshift.io/jenkins-build-uri"],
       "jenkinsLogURL":            ["openshift.io/jenkins-log-url"],
       "jenkinsStatus":            ["openshift.io/jenkins-status-json"],
+      "loggingVisualURL":         ["openshift.io/logging.ui.url"],
```
spadgett: I'd prefer `loggingUIURL` to match the annotation (although I think `openshift.io/logging.ui.hostname` and `loggingUIHostname` would be more accurate, see my comment in the other PR).

You'll need to run `grunt build` to create the dist files

spadgett: See https://github.com/openshift/origin-web-common#contributing for how to run set up grunt

jcantrill:
@spadgett updated

PR #160, opened 2017-08-28
Add watch task to automatically run "deploy" task when files in /src change

benjaminapetersen:
If you `bower link` this repo & intend to do some dev while using `origin-web-console` for the feedback loop, its handy to be able to run `grunt watch` so that the build/dist is automatically run on changes.

@spadgett @jeff-phillips-18

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -21,6 +21,15 @@ module.exports = function (grunt) {
           }
         }
       },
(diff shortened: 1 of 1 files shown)


benjaminapetersen:
For reference:
https://github.com/openshift/origin-web-console/pull/1957

PR #170, opened 2017-09-05
Update notifications with id and timestamp

benjaminapetersen:
Per [this comment](https://github.com/openshift/origin-web-console/pull/2001/files#r136993063) in [PR 2001](https://github.com/openshift/origin-web-console/pull/2001) in origin-web-console.

@spadgett @jeff-phillips-18

### DIFF
--- a/src/ui-services/notificationsService.js
+++ b/src/ui-services/notificationsService.js
@@ -18,6 +18,8 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function() 
     };
 
     var addNotification = function (notification) {
+      notification.id = notification.id || _.uniqueId('notification-');
+      notification.timestamp = new Date().toISOString();
       if (isNotificationPermanentlyHidden(notification) || isNotificationVisible(notification)) {
         return;
       }
--- a/dist/origin-web-common-ui.js
+++ b/dist/origin-web-common-ui.js
@@ -1250,7 +1250,9 @@ angular.module('openshiftCommonUI')
     return function (type) {
       var status;
 
-      switch(type) {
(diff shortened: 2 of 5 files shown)


src/ui-services/notificationsService.js
```diff
@@ -18,6 +18,9 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
     };
 
     var addNotification = function (notification) {
+      notification.id = notification.id || _.uniqueId('notification-');
+      // This format matches timestamp of API objects
+      notification.timestamp = moment.parseZone(new Date()).utc().format();
```
spadgett: I think just

```js
new Date().toISOString();
```

does what we want, then we don't need to add moment.js to common

spadgett: And it's simpler :)

benjaminapetersen: True. lets do that.

benjaminapetersen:
Just updated the filters as well to take the api object type (Normal, Warning).

src/ui-services/notificationsService.js
```diff
@@ -18,6 +18,10 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
     };
 
     var addNotification = function (notification) {
+      notification.id = notification.id || _.uniqueId('notification-');
+      // This format matches timestamp of API objects
+      //notification.timestamp = moment.parseZone(new Date()).utc().format();
```
spadgett: Remove commented out code

benjaminapetersen: got it

PR #171, opened 2017-09-05
Null out handle in DataService.unwatch

benjaminapetersen:
Per [this comment](https://github.com/openshift/origin-web-console/pull/2001#discussion_r136979769) & after doing the `null` manually in a few places, thinking we can update `DataService` with this little tweak.

@spadgett

### DIFF
--- a/src/services/dataService.js
+++ b/src/services/dataService.js
@@ -772,6 +772,7 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
       this._watchInFlight(key, false);
       this._watchOptions(key, null);
     }
+    handle = null;
   };
 
   // Takes an array of watch handles and unwatches them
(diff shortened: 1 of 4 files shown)


src/services/dataService.js
```diff
@@ -772,6 +772,7 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
       this._watchInFlight(key, false);
       this._watchOptions(key, null);
     }
+    handle = null;
```
spadgett: Hey @benjaminapetersen I don't think it makes much difference here since `handle` is local to the function, so it goes out of scope immediately.

benjaminapetersen: aye let me check that.

spadgett:
Closing as I don't think this is necessary.

PR #174, opened 2017-09-06
Improve notification id

benjaminapetersen:
- _.uniqueId() is not sufficient since the ID is used to persist some data
  for notifications via session storage.  The id generated starts at a low
  number & increments w/o randomness.
- Using _.uniqueId() along with Date.now() adds a random value that cannot
  be replicated across sessions/refresh

Example: with uniqueIds like `notification_1`, `notification_2` it is relatively easy for a notification to be marked `read`, but following a `refresh` having `read` be applied to a completely different notification. 

Needed for Notification Drawer updates [web console PR 2001](https://github.com/openshift/origin-web-console/pull/2001).  

@spadgett @jeff-phillips-18

### DIFF
--- a/src/ui-services/notificationsService.js
+++ b/src/ui-services/notificationsService.js
@@ -18,7 +18,7 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function() 
     };
 
     var addNotification = function (notification) {
-      notification.id = notification.id || _.uniqueId('notification-');
+      notification.id = notification.id || _.uniqueId('notification-') + Date.now();
       notification.timestamp = new Date().toISOString();
       if (isNotificationPermanentlyHidden(notification) || isNotificationVisible(notification)) {
         return;
--- a/dist/origin-web-common-ui.js
+++ b/dist/origin-web-common-ui.js
@@ -2149,7 +2149,7 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function() 
     };
 
     var addNotification = function (notification) {
-      notification.id = notification.id || _.uniqueId('notification-');
+      notification.id = notification.id || _.uniqueId('notification-') + Date.now();
       notification.timestamp = new Date().toISOString();
       if (isNotificationPermanentlyHidden(notification) || isNotificationVisible(notification)) {
         return;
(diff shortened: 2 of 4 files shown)


src/ui-services/notificationsService.js
```diff
@@ -18,7 +18,7 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
     };
 
     var addNotification = function (notification) {
-      notification.id = notification.id || _.uniqueId('notification-');
+      notification.id = notification.id || _.uniqueId('notification-') + Date.now();
```
spadgett: @benjaminapetersen This has already merged, but I realize we can't use `notification.id` if set by the caller for `track by` because it won't always be unique. For some messages, we purposely reuse `notification.id` so they don't get toasted twice.

We probably need to always generated some of unique ID that is set by the service and is always unique. So this might need to be.

```js
notification.uniqueID = _.uniqueId('notification-') + Date.now();
```

Then we can use that always for `track by`. This will also fix an animation issue I've noticed with the toast that the slide in animation is reset for already-visible items when new items are added.

benjaminapetersen: So the notification drawer only needs this `uniqueID`, correct?

I'm wondering if we can rename the filed that gets reused so we don't double toast.  I feel like `id` and `uniqueID` is a bit confusing.  I imagine this creates more work for us elsewhere.  

Thoughts?

spadgett: I'd like to use it for `track by` in the toast notifications as well. We have some animation quirks with the toasts because `track by` isn't correct.

spadgett: Renaming the field will touch dozens of files across all 3 repositories :/

benjaminapetersen: Boo...

So we provide `uniqueID` and always use that for `track by`.  If the originator of the notification does not set `id`, we can safely still use the same generated random `id` for it (like here).  So `id` and `uniqueID` can match if its irrelevant, correct?

spadgett: I don't think it's too terrible if we give the new field a very specific name, maybe`trackByID`, and a comment that it's there for uniqueness. What do you think?

Yeah, we should always use the unique ID for track by so avoid duplicates in a repeater error. I wouldn't worry about setting `notification.id` if the caller didn't set one.

benjaminapetersen: works for me! i can open a quick PR to swap this again. 😄

PR #180, opened 2017-09-08
Update notification id to unique id specifically for  in templates

benjaminapetersen:
Per [this conversation](https://github.com/openshift/origin-web-common/pull/174#discussion_r137827377) on a previous PR.

@spadgett @jeff-phillips-18


benjaminapetersen:
@spadgett yup updated & PR to console coming w/the change to match.

PR #182, opened 2017-09-11
Adopt new service catalog resource names

spadgett:
https://trello.com/c/y1lbFrbp

Instance -> ServiceInstance
Binding -> ServiceInstanceCredential

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -1075,7 +1075,7 @@ angular.module("openshiftCommonServices")
 
     var bindingResource = {
       group: 'servicecatalog.k8s.io',
-      resource: 'bindings'
+      resource: 'serviceinstancecredentials'
     };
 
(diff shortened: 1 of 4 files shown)


spadgett:
Rebased

benjaminapetersen:
Worth a conversation about a follow-on, do we want to update our internal vars & service names as well:

```JavaScript 
// now a credentialResource?
var bindingResource = {
  group: 'servicecatalog.k8s.io',
  resource: 'serviceinstancecredentials'
};		    

// CredentialService?
angular.module("openshiftCommonServices")
  .service("BindingService",

// etc.
```
Could prob make a tech debt issue

spadgett:
Good question. We still call it "Binding" in the UI, and we will eventually make pod presets to bind to apps as well.

benjaminapetersen:
No strong opinion then, unless we update to "Credentials" in the UI at some point.

PR #183, opened 2017-09-12
Add ui config for notifications

benjaminapetersen:
Is it reasonable to consider driving how notifications appear from the notifications themselves?  Events are treated separately, via constants.js.

@spadgett @jeff-phillips-18 

re [console PR 2001](https://github.com/openshift/origin-web-console/pull/2001), and others to follow


src/ui-services/notificationsService.js
```diff
@@ -21,6 +21,10 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
       // notifications may already have an id that is not necessarily unique,
       // this is an explicitly unique id just for `track by` in templates
       notification.trackByID = _.uniqueId('notification-') + Date.now();
```
benjaminapetersen: fallback would cover all the notifications currently. Going forward, we could define `toast:true|| false`, `drawer:true|| false`

dtaylor113: In my WIP I added a notification.showToast true||false to show/hide toast notifications being thrown from adding a notification to the drawer.

spadgett: @benjaminapetersen @dtaylor113 It makes sense to me to be able to control toast and drawer behavior per notification. If the default is to toast and not put it in the drawer, maybe

`notification.skipToast` (default: `false`)
`notification.showInDrawer` (default: `false`)

Opinion?

dtaylor113: Hard to know what the default should be.  For quota notifications, they should go into the drawer, but not toasted.

dtaylor113: Also, should we allow toast: false and drawer: false  -would that result in a lost notification?

benjaminapetersen: @spadgett I'm generally good with that, so long as our iterations on the drawer don't shift our presumptions so much that some time in the future we are confused about which is the default & which is the overridden (thus making out defaults confusing).    

@dtaylor113 we'd prob catch that in review?

spadgett: > I'm generally good with that, so long as our iterations on the drawer don't shift our presumptions so much that some time in the future we are confused about which is the default & which is the overridden (thus making out defaults confusing).

Right that's why I'm suggesting the default be `false` always. If it's missing, it's `false`. Then it's less confusing.

If we changed the default, we'd probably change the property name. Since they're exceptions, it's not likely used in too many places.

benjaminapetersen: @dtaylor113 i guess we could check & `Logger.log` it if a notification is misconfigured.  I'm less concerned about it though, this doesn't seem like something that would slip under the radar.

benjaminapetersen:
Updated

src/ui-services/notificationsService.js
```diff
@@ -21,6 +21,8 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function()
...
+      notification.skipToast = notification.skipToast || false;
+      notification.showInDrawer = notification.showInDrawer || false;
```
dtaylor113: I would think ```showInDrawer``` would default to ```true```?

spadgett: The notifications are formatted very differently from events (icons, titles, and messages), so it can look strange to mix them. For example:

screen shot 2017-09-06 at 1 27 35 pm

benjaminapetersen: Yeah, I wonder if we need to clarify hierarchy.  The ideal would be that the events page retains ALL history (events + notifications), the drawer a curated history (important events + notifications), and toasts be less frequent but timely notifications (and ideally not notifying a user of something they already know, ie, just clicked the "build" button).  This has a few problems in that TTL for notifications and TTL for events is entirely different.  We don't store notifications at all, but events have a 2hr TTL from the server.   We also don't include any notification information on the events list page.

@dtaylor113 we are really hashing this out in discussion [here](https://github.com/openshift/origin-web-console/pull/2001#issuecomment-327587103) and [here](https://github.com/openshift/origin-web-console/pull/2001#issuecomment-327790513) (and everything in between).  The problem is that we are a excessively pestering the user sometimes.  

We should probably keep in mind the "mini-toast" idea, would we need an extra config option here? (suggested mini toast would be like gmail notifications):

screen shot 2017-09-06 at 3 16 27 pm

Thoughts?

src/components/toast-notifications/toast-notifications.html
```diff
@@ -1,5 +1,7 @@
 <div class="toast-notifications-list-pf">
-  <div ng-repeat="(notificationID, notification) in notifications track by notification.trackByID" ng-if="!notification.hidden || notification.isHover"
+  <div
```
benjaminapetersen: Just added `notification.skipToast` to the `ng-if` here.

benjaminapetersen: Work for everyone?  If we good, this is prob ready to roll.

spadgett: Might be better to just not add the notification here:

https://github.com/openshift/origin-web-common/blob/master/src/components/toast-notifications/toastNotifications.js#L59

```js
if (notification.skipToast) {
  return;
}
$scope.$evalAsync(...)
```

dtaylor113:
> Might be better to just not add the notification here:
> 
> https://github.com/openshift/origin-web-common/blob/master/src/components/toast-notifications/toastNotifications.js#L59

Just FYI, that's what I have implemented in my quota branch.

src/components/toast-notifications/toast-notifications.html
benjaminapetersen: Great, updated.  @dtaylor113 you can eliminate that line if you are working off of this PR.

spadgett:
@benjaminapetersen Still WIP or ready?

benjaminapetersen:
ready

PR #186, opened 2017-09-13
Don't show redeploy message if not binding to an application

spadgett:
Don't show a user the message saying they need to redeploy their app if they haven't bound the service directly to an application.

cc @ncameronbritt


ncameronbritt:
Good call, @spadgett. I was actually going to ask you about this.

spadgett:
I hadn't noticed it until just now because I'm not patient enough to wait for the provision to complete...

PR #189, opened 2017-09-18
Use label selectors in ApplicationsService

### LINKED ISSUE #179: Determine if we can use a label filter to exclude old replication controllers
Need to investigate if we can use label filter to exclude old replication controllers in the DataService.list call when retrieving applications. If you're not pruning old deployments, you can potentially have a lot of old replication controllers for previous deployments. It would be nice to exclude these using the API rather than filtering on the client.

spadgett:
Filter out replication controllers and replica sets that have owners using label selectors in calls to `DataService.list`. This is more efficient than filtering on the client.

Fixes #179

### DIFF
--- a/src/services/applicationsService.js
+++ b/src/services/applicationsService.js
@@ -1,7 +1,37 @@
 'use strict';
 
 angular.module("openshiftCommonServices").
-service("ApplicationsService", function($filter, $q, DataService) {
+service("ApplicationsService", function($q, DataService) {
+
+  // List replication controllers in a namespace that are NOT managed by a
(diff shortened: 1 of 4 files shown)


src/services/applicationsService.js
```diff
@@ -3,22 +3,52 @@
 angular.module("openshiftCommonServices").
 service("ApplicationsService", function($filter, $q, DataService) {
```
jeff-phillips-18: $filter is no longer used.

spadgett:
Yes, thanks. Updated

PR #199, opened 2017-09-25
Adopt service catalog API changes

spadgett:
https://trello.com/c/y1lbFrbp

cc @jeff-phillips-18


spadgett:
@jeff-phillips-18 PTAL

spadgett:
Origin PR is in the merge queue.

PR #203, opened 2017-09-26
add ability to "disable" form validity

juanvallejo:
Required by: https://github.com/openshift/origin-web-console/pull/2176

Adds ability to disable form validity based on value of a model.

### DIFF
--- a/src/components/osc-unique/oscUnique.js
+++ b/src/components/osc-unique/oscUnique.js
@@ -31,11 +31,13 @@ angular.module('openshiftCommonUI')
     return {
       restrict: 'A',
       scope: {
-        oscUnique: '='
+        oscUnique: '=',
+        oscUniqueDisabled: '='
       },
       require: 'ngModel',
       link: function($scope, $elem, $attrs, ctrl) {
         var list = [];
+        var isUnique = true;
 
         $scope.$watchCollection('oscUnique', function(newVal) {
           list = _.isArray(newVal) ?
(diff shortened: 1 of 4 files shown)


src/components/osc-unique/oscUnique.js
```diff
@@ -31,21 +31,41 @@ angular.module('openshiftCommonUI')
...
       scope: {
-        oscUnique: '='
+        oscUnique: '=',
+        oscUniqueDisabled: '='
       },
       require: 'ngModel',
       link: function($scope, $elem, $attrs, ctrl) {
         var list = [];
+        var disabled = false;
+        var isValid = true;
```
spadgett: I suggest `isUnique` here. It's technically valid if disabled even when not unique, so I think `isUnique` is more accurate.

src/components/osc-unique/oscUnique.js
```diff
@@ -31,21 +31,41 @@ angular.module('openshiftCommonUI')
...
+          ctrl.$setValidity('oscUnique', disabled);
+        });
+
         ctrl.$parsers.unshift(function(value) {
+          isValid = !_.includes(list, value);
+
+          // if disabled, skip validation, return true
+          if(disabled) {
+            ctrl.$setValidity('oscUnique', disabled);
+            return disabled;
```
spadgett: Isn't this supposed to return the input value, not `true`?

https://docs.angularjs.org/api/ng/type/ngModel.NgModelController#$parsers

I think this can be

```js
ctrl.$parsers.unshift(function(value) {
  isValid = !_.includes(list, value);
  ctrl.$setValidity('oscUnique', $scope.oscUniqueDisabled || isValid);

  return value;
});
```

You could pull the `$setValidity` line into its own function above so it's not repeated.

```js
var isUnique = true;

var updateValiditity = function() {
  ctrl.$setValidity('oscUnique', $scope.oscUniqueDisabled || isUnique);
};
$scope.$watch('oscUniqueDisabled', updateValidity);

ctrl.$parsers.unshift(function(value) {
  isUnique = !_.includes(list, value);
  updateValidity();
  return value;
});
```

juanvallejo:
@spadgett thanks for the feedback, review comments addressed

PR #205, opened 2017-09-27
[WIP] Bumping version to v0.0.61

rhamilto:


spadgett:
Closing since we released a 0.0.61 yesterday evening.

@rhamilto FYI, there should be a change in package-lock.json as well when we bump a version, but might require you install a newer npm.

PR #211, opened 2017-10-03
Improving display of binding status message

rhamilto:
localhost-9000-dev-console- iphone 6 1
localhost-9000-dev-console- iphone 6 2

screen shot 2017-10-03 at 12 12 52 pm
screen shot 2017-10-03 at 12 13 10 pm


src/components/binding/bindResults.html
```diff
@@ -1,31 +1,36 @@
 <div ng-if="!ctrl.error && !(ctrl.binding | isBindingFailed)">
   <div ng-if="ctrl.binding && !(ctrl.binding | isBindingReady)" class="results-status">
-    <span class="spinner spinner-sm" aria-hidden="true"></span>
+    <span class="fa fa-clock-o text-muted" aria-hidden="true"></span>
```
jeff-phillips-18: The spinner was part of the UX design. I like this better but should be approved by @openshift/team-ux-review

src/styles/_bind-service.less
```diff
@@ -52,23 +52,49 @@
...
+    @media(min-width: @screen-sm-min) {
+      font-size: 20px;
+      font-weight: 300;
+    }
+    strong {
+      font-weight: 600;
+      .word-break();
+    }
+  }
+}
```
jeff-phillips-18: Seems odd to me to use h3 and then override almost all the values. Does this have to be an h3?

src/components/binding/bindResults.html
rhamilto: I've been working with @ncameronbritt and co. on this.  The clock is the approved solution.

src/styles/_bind-service.less
rhamilto: The existing node was an h3, and I didn't change that.  I believe it should be a heading of some value so it has semantic value (probably an h1).  It looks like you originally authored it as an h3.  Any insight as to why?  Heading styles can and should be overridden.  Semantics !== style.

jeff-phillips-18: I don't recollect why it was an h3. I'm ok with this as is. Thanks.

spadgett:
Tests ran cleanly locally:

> PhantomJS 2.1.1 (Mac OS X): Executed 153 of 153 SUCCESS (0.59 secs / 0.67 secs)

PR #215, opened 2017-10-05
Added isNotificationHidden to NotificationsService

dtaylor113:
This is needed for https://bugzilla.redhat.com/show_bug.cgi?id=1495455#c4

`isNotificationHidden` returns `true` is user specifically cleared/hid/perm. hid the notification in the notification drawer, else returns `false`.

Also added `namespace` awareness to `isNotificationVisible` and `hideNotification` functions.

### DIFF
--- a/dist/origin-web-common-ui.js
+++ b/dist/origin-web-common-ui.js
@@ -2227,6 +2227,18 @@ angular.module('openshiftCommonUI').provider('NotificationsService', function() 
       _.take(notifications, 0);
     };
 
+    // Was the notification set to be hidden?
+    // This is different than !isNotificationVisible
+    var isNotificationHidden = function (notification) {
+      if (!notification.id) {
+        return false;
+      }
+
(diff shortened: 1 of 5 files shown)


spadgett:
I guess I'm unclear why this is needed. I'd rather not have to set a `namespace` property on notifications everywhere.

I was assuming the notification drawer can just include the namespace in the ID it uses for tracking whether a notification is cleared since it knows the current namespace.

dtaylor113:
> I was assuming the notification drawer can just include the namespace in the ID

Oh, ok, that is the way I originally had it, but I thought your email indicated that you wanted the NotificationsService to take into account a `namespace`.    I'll adjust, thanks

dtaylor113:
> I'd rather not have to set a namespace property on notifications everywhere.

Ok, fixed.

dtaylor113:
Not needed

PR #216, opened 2017-10-05
[WIP] EnvFrom Styling

cdcabrera:
Styling updates for EnvFrom component. 

- Minor select list highlighting update. Related to active and disabled classes being applied to the first UI-Select option under the scenario of it being auto selected.
- [WIP] Modal styling classes

### DIFF
--- a/src/styles/_ui-select.less
+++ b/src/styles/_ui-select.less
@@ -18,7 +18,8 @@
       > span {
         background-color: @dropdown-link-active-bg !important;
         border-color: @dropdown-link-active-border-color !important;
-        color: @dropdown-link-active-color;
+        color: @dropdown-link-active-color !important;
+        cursor: default !important;
       }
       .text-muted {
         color: #f5f5f5;
--- a/dist/less/_ui-select.less
+++ b/dist/less/_ui-select.less
@@ -18,7 +18,8 @@
       > span {
         background-color: @dropdown-link-active-bg !important;
         border-color: @dropdown-link-active-border-color !important;
-        color: @dropdown-link-active-color;
+        color: @dropdown-link-active-color !important;
+        cursor: default !important;
       }
       .text-muted {
         color: #f5f5f5;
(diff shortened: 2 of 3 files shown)


src/styles/_ui-select.less
```diff
@@ -18,7 +18,8 @@
       > span {
         background-color: @dropdown-link-active-bg !important;
         border-color: @dropdown-link-active-border-color !important;
-        color: @dropdown-link-active-color;
+        color: @dropdown-link-active-color !important;
+        cursor: default !important;
```
spadgett: Wouldn't be want this to be `pointer`?

cdcabrera: So this is the funky part. This scenario plays out as that item being disabled... we could do a few things... 
- leave it as the disabled cursor, 
- change it to the pointer to make it similar to the select lists in general, 
- or leave it as something neutral with default...

sg00dwin: This screenshot shows the conundrum and the reason why, to me, it made more sense to assign it the `default` cursor.

By default ui-select makes the first selection .active. So if the first option as already been selected in another select, then having the cursor be `pointer` would imply it’s clickable when it’s not.


screen shot 2017-10-05 at 4 40 34 pm

spadgett: But it should be clickable. The bug to me is that the current value is marked invalid. It is a valid selection.

Note that in other selects we use a pointer cursor for the active item.

spadgett: In other words, I don't think this is a style bug. The bug is we're making an item disabled when it's not.

cdcabrera: We agree it's not entirely a style bug when an item can be disabled and active at the same time... but at the same time UI Select maybe should have already taken that into account with it's styling... or lack of.

Kind of the opinion that having both classes applied really shouldn't be able to happen without some form of CSS support if UI Select is just going to default to selecting the first item automatically.

Working on a couple of alternates:
- we just ignore disabling all together
- @sg00dwin mentioned an alternative that has it's own issues (possibly around drag and sort on the Environment From component) where we could simply remove/hide the items already selected... it avoids the issue entirely.
- we try to autoselect the first non-disabled item

cdcabrera:
Closing this PR since it's related to a bypass as opposed to a real fix. May re-open as necessary. 

[Origin Web Console PR EnvFrom Config Map and Secret Link 2201 contains the approach of relaxing the disabled aspect and placing an error/warning message instead](https://github.com/openshift/origin-web-console/pull/2201)

PR #218, opened 2017-10-05
Decorate links with a target with fa-external-link

### LINKED ISSUE #217: linkify with target should include external link icon in output
We typically include fa-external-link with links that include target="_blank".  For consistency, it seems we would want linkify to behave the same way.  @spadgett, agree/disagree?

rhamilto:
Fixes #217

### DIFF
--- a/src/ui-services/htmlService.js
+++ b/src/ui-services/htmlService.js
@@ -102,7 +102,7 @@ angular.module("openshiftCommonUI")
         // Replace any URLs with links.
         return text.replace(/https?:\/\/[A-Za-z0-9._%+-]+\S*[^\s.;,(){}<>"\u201d\u2019]/gm, function(str) {
           if (target) {
-            return "<a href=\"" + str + "\" target=\"" + target + "\">" + str + "</a>";
+            return "<a href=\"" + str + "\" target=\"" + target + "\">" + str + " <i class=\"fa fa-external-link\" aria-hidden=\"true\"></i></a>";
           }
 
           return "<a href=\"" + str + "\">" + str + "</a>";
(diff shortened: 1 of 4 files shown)


src/ui-services/htmlService.js
```diff
@@ -102,7 +102,7 @@ angular.module("openshiftCommonUI")
         // Replace any URLs with links.
         return text.replace(/https?:\/\/[A-Za-z0-9._%+-]+\S*[^\s.;,(){}<>"\u201d\u2019]/gm, function(str) {
           if (target) {
-            return "<a href=\"" + str + "\" target=\"" + target + "\">" + str + "</a>";
+            return "<a href=\"" + str + "\" target=\"" + target + "\">" + str + " " + "<i class=\"fa fa-external-link\" aria-hidden=\"true\"></i></a>";
```
spadgett: nit: You have an extra `+ " " +` that you don't need.

```js
return "<a href=\"" + str + "\" target=\"" + target + "\">" + str + " <i class=\"fa fa-external-link\" aria-hidden=\"true\"></i></a>";
```

rhamilto: My testing indicates we do need the extra space or the icon is butted up against the text.  :/

spadgett: You can have the space, but still make it one string literal

```
" <i class=\"...\">"
```

Although I wonder if it shouldn't be an `&nbsp;` or `&thinsp;`?

rhamilto: > You can have the space, but still make it one string literal

That works.  I had tried that before and it didn't work for some reason.  Probably error on my part.  :)

rhamilto:
@spadgett, if you're good with this, it's ready for merge.

PR #223, opened 2017-10-09
Adopt more service catalog API changes

spadgett:
* Cluster prefix for service classes and plans
* ServiceInstanceCredential -> ServiceBinding

cc @jeff-phillips-18


jeff-phillips-18:
LGTM

PR #230, opened 2017-10-16
Bug 1502574 - Don't fail on all discovery errors

spadgett:
Some errors are not fatal for API groups the console doesn't use.
We should log warnings for those errors and continue.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1502574
Fixes https://github.com/openshift/origin-web-console/issues/2014

### DIFF
--- a/src/openshiftCommonServices.module.js
+++ b/src/openshiftCommonServices.module.js
@@ -54,7 +54,8 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
       API_DISCOVERY_ERRORS.push({
         data: data,
         textStatus: textStatus,
-        xhr: jqXHR
+        xhr: jqXHR,
+        fatal: true
       });
     });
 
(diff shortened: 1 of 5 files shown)


src/openshiftCommonServices.module.js
```diff
@@ -128,26 +130,28 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
...
+  //   .then(_.partial(getGroups, baseURL, server), function(data, textStatus, jqXHR) {
+  //     if (server.required !== false) {
+  //       API_DISCOVERY_ERRORS.push({
+  //         data: data,
+  //         textStatus: textStatus,
+  //         xhr: jqXHR
+  //       });
+  //     }
+  //   }));
+  // });
```
jeff-phillips-18: Any reason to hang onto this code?

spadgett: I can yank it. Wasn't sure if it would be useful in the future, but it's always in the git history.

spadgett:
@jeff-phillips-18 thanks, updated

PR #239, opened 2017-10-26
Fix bugs sorting service instances

spadgett:
* Fix call to `_.sortBy` so that we're correctly handling secondary sort.
* Add utility that always waits for all promises to complete (even if one has failed).

See https://github.com/openshift/origin-web-console/issues/2384
See https://github.com/openshift/origin-web-console/pull/2385

### DIFF
--- a/src/services/bindService.js
+++ b/src/services/bindService.js
@@ -188,19 +188,12 @@ angular.module("openshiftCommonServices")
     };
 
     var sortServiceInstances = function(serviceInstances, serviceClasses) {
-      if (!serviceInstances && !serviceClasses) {
-        return null;
-      }
+      var getServiceClassDisplayName = function(serviceInstance) {
+        var serviceClassName = _.get(serviceInstance, 'spec.clusterServiceClassRef.name');
+        return _.get(serviceClasses, [serviceClassName, 'spec', 'externalMetadata', 'displayName']) || serviceInstance.spec.clusterServiceClassExternalName;
+      };
 
-      return _.sortBy(serviceInstances,
-        function(item) {
-          var serviceClassName = _.get(item, 'spec.clusterServiceClassRef.name');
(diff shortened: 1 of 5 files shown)


src/services/bindService.js
```diff
@@ -188,19 +188,12 @@ angular.module("openshiftCommonServices")
     };
 
     var sortServiceInstances = function(serviceInstances, serviceClasses) {
-      if (!serviceInstances && !serviceClasses) {
-        return null;
-      }
+      var getServiceClassDisplayName = function(serviceInstance) {
+        var serviceClassName = _.get(serviceInstance, 'spec.clusterServiceClassRef.name');
+        return _.get(serviceClasses, [serviceClassName, 'spec', 'externalMetadata', 'displayName']) || serviceInstance.spec.clusterServiceClassExternalName;
```
jwforres: just checking, external name is required right?

src/services/promiseUtilsService.js
```diff
@@ -0,0 +1,51 @@
...
+        var deferred = $q.defer();
+        var complete = 0;
+        var failed = false;
+        var checkDone = function() {
+          if (complete < total) {
+            return;
+          }
+
+          if (failed) {
+            deferred.reject();
```
jwforres: i think $q.all passes up the content from the failure when it does a reject, not positive.  Not that we necessarily need that, but we should document above if it is a difference.

src/services/bindService.js
spadgett: I'm not sure it is, but it should be OK if it's not set?

screen shot 2017-10-26 at 12 47 01 pm

src/services/promiseUtilsService.js
spadgett: Thanks, updated the comment.

PR #243, opened 2017-11-03
Position so ui-select overlays when appended to body.

sg00dwin:
Fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1479114


rhamilto:
> lgtm pending confirmation that this was tested with ui-select inside a dialog

Yep.  That's the exact case @sg00dwin is fixing.

PR #248, opened 2017-11-06
Dedupe rbac roles in apiService

benjaminapetersen:
Updates the blacklist in apiService to drop the `authorization.openshift.io` roles (now in `rbac.authorization.k8s.io`).

Will fix [2446](https://github.com/openshift/origin-web-console/issues/2446) in web-console when we vendor it in.  

Tests updated.

@jwforres @spadgett


benjaminapetersen:
rebasing.

dist/origin-web-common-services.js
```diff
@@ -467,6 +467,18 @@ angular.module('openshiftCommonServices')
     return "The API version "+version+" for kind " + kind + " is not supported by this server";
   };
 
+  // Exclude duplicate kinds we know about that map to the same storage as another
+  //  group/kind.  This is unusual, so we are special casing these.
+  var dedupeGroups = [{group: 'authorization.openshift.io'}];
+  var dedupeKinds = [{group: 'extensions', kind: 'HorizontalPodAutoscaler'}];
+
+  var excludeKindFromAPIGroupList = function(groupName, resourceKind) {
```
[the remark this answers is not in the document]
jwforres: calculateAvailableKinds is only called twice during the page load, after that its kept in memory, so as long as its not horribly inefficient im not worried.  This seems fine.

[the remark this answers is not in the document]
benjaminapetersen: Ok cool. may be good to go here then.

test/spec/services/apiServiceSpec.js
```diff
@@ -118,20 +118,57 @@ describe("APIService", function() {
...
+        it('should return the correct verbs', function() {
+          var actualAPIInfo = APIService.apiInfo(rgv);
+          if(actualAPIInfo) {
+            expect(actualAPIInfo.verbs).toEqual(expectedAPIInfo.verbs);
+          }
+        });
+        // TODO: Is this valid or a bad test?
+        // I don't have any idea why it would be undefined.
+        // If its correct, we should keep it and clarify the condition.
+        // it('should be undefined for an unknown reason', function() {
```
jwforres: The case at L116
I believe the point is that its bogus api information that won't resolve to an actual type, so the call to get the api info should return undefined/null

benjaminapetersen: ok makes sense.  updated.

benjaminapetersen:
updated

test/spec/services/apiServiceSpec.js
```diff
@@ -118,21 +118,58 @@ describe("APIService", function() {
...
+        });
+        it('should return the correct verbs', function() {
+          var actualAPIInfo = APIService.apiInfo(rgv);
+          if(actualAPIInfo) {
+            expect(actualAPIInfo.verbs).toEqual(expectedAPIInfo.verbs);
+          }
+        });
     }));
+
+    it('should return undefined if the rgb is unknown', function() {
```
jwforres: if you are making this a separate test case, remove the case in L115-117, it won't get tested by your checks above

benjaminapetersen: Yeah, i thought that 115 just looked too much like a real resource, so the purpose of the test was obscured (given I couldn't easily diagnose).

Eliminating.

PR #250, opened 2017-11-15
Add deploymentconfigs/log to constants/apiPreferredVersions

benjaminapetersen:
@jwforres @spadgett @jeff-phillips-18


benjaminapetersen:
Great Q.  I assume it is because we wanted to be explicit with everything (in case the preferred versions from the APIs ever change).  

I don't have strong feelings on this, perhaps @spadgett does?

spadgett:
No strong opinion. I'm OK either way.

You might add `builds/log` and `pods/log` too while you're in the file.

dist/origin-web-common-services.js
```diff
@@ -3755,6 +3755,8 @@ angular.module('openshiftCommonServices')
       deploymentconfigs:                {group: 'apps.openshift.io',          version: 'v1',      resource: 'deploymentconfigs' },
 	    'deploymentconfigs/instantiate':  {group: 'apps.openshift.io',          version: 'v1',      resource: 'deploymentconfigs/instantiate' },
```
spadgett: nit: mind fixing the indentation here?

benjaminapetersen: in the `dist`, weird.  `src` is good, lemme rebuild & see whats up.

benjaminapetersen:
@spadgett @jwforres any concerns merging this so I can get those other PRs to console merged?  I can always open a new PR to common if more items come up, but I'd rather not carry all those little PRs along if we don't need to.

jwforres:
im ok with this for now, but if we find we are getting a ton of these subresources in here we might want to consider defaulting to the parent resource's version as defined in our list

benjaminapetersen:
Yeah that makes sense, would be easy to modify.

PR #260, opened 2017-12-04
ProjectsService:  Added 'opts.skipErrorNotFound' handling

dtaylor113:
For use in "User Set Home Page Pref.":  https://trello.com/c/sg1xw7gG

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -3169,7 +3169,7 @@ angular.module('openshiftCommonServices')
       };
 
       return {
-        get: function(projectName) {
+        get: function(projectName, opts) {
           return  AuthService
                     .withUser()
                     .then(function() {
@@ -3197,6 +3197,9 @@ angular.module('openshiftCommonServices')
                                         });
                               }, function(e) {
                                 context.projectPromise.reject(e);
+                                if ((e.status === 403 || e.status === 404) && _.get(opts, 'skipErrorNotFound')) {
+                                  return $q.reject({notFound: true});
+                                }
                                 var description = 'The project could not be loaded.';
                                 var type = 'error';
                                 if(e.status === 403) {
--- a/src/services/projects.js
+++ b/src/services/projects.js
@@ -42,7 +42,7 @@ angular.module('openshiftCommonServices')
       };
 
       return {
-        get: function(projectName) {
+        get: function(projectName, opts) {
           return  AuthService
                     .withUser()
                     .then(function() {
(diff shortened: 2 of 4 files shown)


dtaylor113:
@spadgett Hi, can you review and merge if you feel it's ready.  I'm weary of merging code, not sure if you or @jwforres need to approve first?

src/services/authService.js
```diff
@@ -13,6 +13,8 @@ angular.module('openshiftCommonServices')
 //     sets the current user and token to use for authenticated requests
 //     if ttl is specified, it indicates how many seconds the user and token are valid
 //     triggers onUserChanged callbacks if the new user is different than the current user
+//   getUser()
+//     returns the user
```
spadgett: You should explain how this is different from `withUser` (`withUser` triggers a login if there is no user, whereas this method will return null or undefined.)

dtaylor113: Ok, added comments

dist/origin-web-common-services.js
```diff
@@ -3193,6 +3194,9 @@ angular.module('openshiftCommonServices')
                                         });
                               }, function(e) {
                                 context.projectPromise.reject(e);
+                                if ((e.status === 403 || e.status === 404) && _.has(opts, 'skipErrorOnProjNotFound')) {
```
spadgett: `_.has` is wrong here since it will return true even when `skipErrorOnProjNotFound === false`. Use `_.get`.

spadgett: I'd probably go with the shorter `skipErrorNotFound`

dist/origin-web-common-services.js
```diff
@@ -3193,6 +3194,9 @@ angular.module('openshiftCommonServices')
...
+                                  return $q.reject({projNotFound: true});
```
spadgett: Let's go with either `notFound: true` or `projectNotFound: true`. (I have a slight preference for the first.)

dist/origin-web-common-services.js
benjaminapetersen: Would we be interested in a `service/filter/helper` for the `=== 404` kind of logic? Something like `notFound(e) / forbidden(e)`?  I bet we have this scattered around the code.

benjaminapetersen: `opts.error.notFound`?

dtaylor113: Fixed to 'opts.skipErrorNotFound'

dist/origin-web-common-services.js
dtaylor113: Fixed to 'notFound: true'

src/services/projects.js
```diff
@@ -39,7 +39,8 @@ angular.module('openshiftCommonServices')
       };
 
       return {
-        get: function(projectName) {
+        get: function(projectName, opts) {
+          opts = opts || {};
```
spadgett: nit: you can remove this line since you use `_.get` below, which handles undefined values

...or you can leave this line and check the property directly below without `_.get`

```js
if ((e.status === 403 || e.status === 404) && opts.skipErrorNotFound) {
```

dtaylor113: fixed, removed line

PR #263, opened 2017-12-08
Update rbac to v1beta1 in API_PREFERRED_VERSIONS

benjaminapetersen:
Gotta bump the version up a notch for membership to work


spadgett:
> Gotta bump the version up a notch for membership to work

So I'm confused by this. Isn't `v1beta1` older than `v1`? What's not working with `v1`?

@jwforres @enj

spadgett:
We already merged https://github.com/openshift/origin-web-console/pull/2550 without this change. @benjaminapetersen Do you remember what problem you were seeing?

benjaminapetersen:
Thinking. I spoke with @enj about this one, but I don't remember off the top of my head.  Fishing around now to see if I can figure it out.  Sorry I didn't leave a better description.

benjaminapetersen:
I'm going to close this, the membership page is working as expected for me... I'll reopen if I can create a bug that leads me to think this is still relevant.

enj:
This will work in 3.9+, it will fail in 3.7-

spadgett:
The console changes are only in 3.9, so we should be OK.

benjaminapetersen:
Will we need to worry more about these changes once we run in a pod?  Perhaps have a way to match up the versions?

jwforres:
my only concern would be the 3.7 to 3.9 upgrade, when we are in the middle of switching over the masters you could have a 3.9 console hitting a 3.7 master, this would have been true even if we weren't switching to running in a pod

jwforres:
@enj is the issue that the v1 API doesn't exist in 3.7? @benjaminapetersen @spadgett we probably need the preferred version logic to fall back to the server's preferred version if our preferred version isn't returned by API discovery.  This would let us prefer newer versions but still be backward compat during upgrades.

benjaminapetersen:
How worried are we that the objects themselves change enough that code would break, in this case?

jwforres:
I think we need to have a bigger discussion about this, @spadgett @benjaminapetersen we should talk tomorrow.

This may be the first time we have actually hit this issue, but just recently we said we were going to make an effort to be n-1 compatible from console to APIs to prevent breakages during upgrades.

If we are jumping major API versions, or we know we are moving from beta API to GA API then we need to be aware of the differences and not break.

How big of a change are we talking about in this RBAC case, are there actually differences between v1beta1 and v1?

spadgett:
@enj is `v1` there in 3.8?

spadgett:
If `v1` is in 3.8, the 3.7 -> 3.8 -> 3.9 upgrade that happens should work.

jwforres:
Now I feel like we had this same conversation 3 weeks ago 🤔

enj:
> my only concern would be the 3.7 to 3.9 upgrade, when we are in the middle of switching over the masters you could have a 3.9 console hitting a 3.7 master, this would have been true even if we weren't switching to running in a pod

Yup, rolling upgrades.

> @enj is the issue that the v1 API doesn't exist in 3.7?

Correct.

> How worried are we that the objects themselves change enough that code would break, in this case?

These objects are stable, and will never have any backwards incompatible changes made to them.  `v2alpha1` would start any such changes.

> How big of a change are we talking about in this RBAC case, are there actually differences between v1beta1 and v1?

`v1beta1` and `v1` for the RBAC types are 100% identical, minus the addition of the optional `AggregationRule` field in cluster role in `v1`.  This has no effect on the web console.

> @enj is `v1` there in 3.8?

Yes.

---

@spadgett @jwforres @benjaminapetersen why not use discovery to determine the safest version to use?  For example, any 3.9 client that wants to be `n-1` compatible should default to `v1beta1` unless it knows the server supports `v1` (rolling upgrades can still cause issues based on who you performed discovery against, so `v1beta1` is the safer choice).

Origin has had to deal with the rolling upgrade problem since the beginning.  For example, even though 3.8 supports `v1`, it stores things internally as `v1beta1` to make sure an older master can still read the data.  We have tests that assert this for all types: https://github.com/openshift/origin/blob/release-3.8/test/integration/etcd_storage_path_test.go#L638-L642

benjaminapetersen:
I think if we are going to start supporting multiple API versions we will definitely want a test strategy.

PR #267, opened 2017-12-08
Add oauthaccesstokens to API_PREFERRED_VERSIONS, update deleteTokenLogoutService

benjaminapetersen:
- APIService in deleteTokenLogoutService would create a circular dependency

### DIFF
--- a/src/constants/apiPreferredVersions.js
+++ b/src/constants/apiPreferredVersions.js
@@ -25,6 +25,7 @@ angular.module('openshiftCommonServices')
       imagestreamimages:                {group: 'image.openshift.io',         version: 'v1',      resource: 'imagestreamimages' },
       imagestreamimports:               {group: 'image.openshift.io',         version: 'v1',      resource: 'imagestreamimports' },
       limitranges:                      {version: 'v1',                       resource: 'limitranges' },
+      oauthaccesstokens:                {group: 'oauth.openshift.io',         version: 'v1',      resource: 'oauthaccesstokens' },
       pods:                             {version: 'v1',                       resource: 'pods' },
       'pods/log':                       {version: 'v1',                       resource: 'pods/log' },
       processedtemplates:               {group: 'template.openshift.io',      version: 'v1',      resource: 'processedtemplates' },
--- a/src/services/deleteTokenLogoutService.js
+++ b/src/services/deleteTokenLogoutService.js
@@ -4,9 +4,11 @@
 angular.module('openshiftCommonServices')
 .provider('DeleteTokenLogoutService', function() {
 
-  this.$get = function($q, $injector, Logger) {
+  this.$get = function(
+    $q,
+    $injector,
+    Logger) {
     var authLogger = Logger.get("auth");
-
     return {
       logout: function(user, token) {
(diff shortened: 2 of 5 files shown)


src/services/deleteTokenLogoutService.js
```diff
@@ -6,7 +6,8 @@ angular.module('openshiftCommonServices')
 
   this.$get = function($q, $injector, Logger) {
     var authLogger = Logger.get("auth");
-
+    // Using APIService creates a circular dependency
```
spadgett: Does this work?

```js
var APIService = $injector.get("APIService");
```

benjaminapetersen: Yeah good call, like `DataService` below, updating

benjaminapetersen:
updated

src/services/deleteTokenLogoutService.js
```diff
@@ -19,11 +21,12 @@ angular.module('openshiftCommonServices')
 
         // Lazily get the data service. Can't explicitly depend on it or we get circular dependencies.
         var DataService = $injector.get('DataService');
+        var oAuthAccessTokensVersion = $injector.get('APIService').getPreferredVersion('oauthaccesstokens');
```
spadgett: Nit: I'd find this hard to read as one statement. I'd prefer something like...

```js
// Lazily get the API service and data service. Can't explicitly depend on them or we get circular dependencies.
var APIService = $injector.get('APIService');
var DataService = $injector.get('DataService');

var oAuthAccessTokensVersion = APIService.getPreferredVersion('oauthaccesstokens');
```

benjaminapetersen: I'm fine with that. updating.

benjaminapetersen:
rebased

src/services/deleteTokenLogoutService.js
```diff
@@ -18,12 +20,14 @@ angular.module('openshiftCommonServices')
         }
 
         // Lazily get the data service. Can't explicitly depend on it or we get circular dependencies.
```
spadgett: do you mind updating the comment since it applies to both services?

benjaminapetersen: Sure thing!

benjaminapetersen:
updated

src/constants/apiPreferredVersions.js
```diff
@@ -25,6 +25,7 @@ angular.module('openshiftCommonServices')
       imagestreamimages:                {group: 'image.openshift.io',         version: 'v1',      resource: 'imagestreamimages' },
       imagestreamimports:               {group: 'image.openshift.io',         version: 'v1',      resource: 'imagestreamimports' },
       limitranges:                      {version: 'v1',                       resource: 'limitranges' },
+      oauthaccesstokens:                {version: 'v1',                       resource: 'oauthaccesstokens' },
```
spadgett: `group: oauth.openshift.io/v1`

benjaminapetersen: ah, thx. got it.

benjaminapetersen:
Think we are good on this one.

PR #268, opened 2017-12-14
Add toAPIVersion method to APIService

benjaminapetersen:


dist/origin-web-common-services.js
```diff
@@ -260,6 +260,15 @@ angular.module('openshiftCommonServices')
     return new ResourceGroupVersion(resource, group, version);
   };
 
+  // toGroupVersion() returns a group/version string
+  //
+  var toGroupVersion = function(resourceGroupVersion) {
```
benjaminapetersen: I'm trying to decide if this name is ok, given that the other fns with similar names return objects.

jwforres: the appropriate name is toAPIVersion I think, since that is the field in the API objects that it refers to

spadgett: +1 for `toAPIVersion`

benjaminapetersen: perfect, updating.

benjaminapetersen:
updated

PR #270, opened 2017-12-15
Add normalizeIconClass filter

rhamilto:
Supports https://github.com/openshift/origin/issues/17745

Required by https://github.com/openshift/origin-web-catalog/pull/603 and https://github.com/openshift/origin-web-console/pull/2607

### DIFF
--- a/src/filters/normalizeIconClass.js
+++ b/src/filters/normalizeIconClass.js
@@ -1,0 +1,14 @@
+'use strict';
+
+angular.module("openshiftCommonUI")
+  .filter("normalizeIconClass", function() {
+    return function(iconClass) {
+      // if iconClass starts with "icon-", append "font-icon "
+      // so the Openshift Logos Icon font is used
+      if(_.startsWith(iconClass, "icon-")) {
+        return "font-icon " + iconClass;
+      } else {
+        return iconClass;
+      }
+    };
+  });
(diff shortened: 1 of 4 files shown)


src/filters/normalizeIconClass.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+angular.module("openshiftCommonUI")
+  .filter("normalizeIconClass", function() {
+    return function(iconClass) {
+      // if iconClass starts with "icon", append "icon-font "
+      // so the Openshift Logos Icon font is used
+      if(_.startsWith(iconClass, "icon")) {
```
spadgett: Should we be checking for "icon" or "icon-"? I believe all of the OpenShift logos start with "icon-"

src/filters/normalizeIconClass.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+angular.module("openshiftCommonUI")
+  .filter("normalizeIconClass", function() {
+    return function(iconClass) {
+      // if iconClass starts with "icon", append "icon-font "
```
spadgett: s/icon-font/font-icon

src/filters/normalizeIconClass.js
spadgett: It might be good to keep the hyphen to avoid false positives

rhamilto:
Good catches, @spadgett.  Fixed.

PR #274, opened 2018-01-03
allow logging index for a project to be specified via annotation

jcantrill:
This PR is an annotation for defining where to find user log data to resolve bz1523047


jcantrill:
@spadgett How do I cherrypick this to resolve https://bugzilla.redhat.com/show_bug.cgi?id=1547348 for release-3.7

spadgett:
I’d use the full annotation name directly in console and skip this commit. Otherwise we need to cut a new common release. I don’t think it’s worth it

PR #281, opened 2018-01-22
Switch karma browser to Firefox for Travis

TheRealJon:
Fix recent build failure issue by using Firefox browser in Karma when building on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ cache:
   directories:
   - node_modules
   - bower_components
-addons:
-  chrome: stable
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -77,7 +77,7 @@ module.exports = function(config) {
     // - PhantomJS
     // - IE (only Windows)
     // CLI --browsers Chrome,Firefox,Safari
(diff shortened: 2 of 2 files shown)


jwforres:
would be curious if you kick this in travis a bunch of times if you can ever get it to flake like chrome does

spadgett:
We should remove the chrome stuff from .travis.yml if it's not needed.

https://github.com/openshift/origin-web-common/blob/master/.travis.yml#L8-L9

dtaylor113 (approved):

TheRealJon:
@spadgett Removed the chrome addon from the travis config.
@jwforres I've run the build several times on my fork and it seems to be consistently passing. It has also run about 3 times on this PR without issue.

spadgett (dismissed):

spadgett:
@TheRealJon It looks like the tests aren't actually running:

> Firefox 56.0.0 (Linux): Executed 0 of 386 SUCCESS (0 secs / 0 secs)

TheRealJon:
@spadgett It looks like that is just the original state. Looking further down the log, this status gets updated periodically.

https://travis-ci.org/openshift/origin-web-common/builds/331966815#L1578
https://travis-ci.org/openshift/origin-web-common/builds/331966815#L1838
etc...

spadgett:
Oops I think you're right. Seeing `0 secs` all the way down made me think they didn't run.

spadgett (approved):

PR #284, opened 2018-01-23
URI path validation pattern

jhadvig:
@jwforres PTAL

### DIFF
--- a/src/openshiftCommonUI.module.js
+++ b/src/openshiftCommonUI.module.js
@@ -22,6 +22,14 @@ angular.module('openshiftCommonUI', [])
   maxlength: 253,
   description: 'Name must consist of lower-case letters, numbers, periods, and hyphens. It must start and end with a letter or a number.'
 })
+// URI_PATH_FRAGMENT_VALIDATION patterns must consist of following characters:
+// a-z A-Z 0-9 . - _ ~ ! $ & ' ( ) * + , ; = : @
+// https://stackoverflow.com/questions/4669692/valid-characters-for-directory-part-of-a-url-for-short-links
+.constant('URI_PATH_FRAGMENT_VALIDATION', {
+  pattern: /^[a-zA-Z0-9.\-_~!$&'()*+,;=:@]+$/,
+  minlength: 8,
+  description: 'Value must consist of characters that can be contained URI path segment.'
+})
 // http://stackoverflow.com/questions/9038625/detect-if-device-is-ios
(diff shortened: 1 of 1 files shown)


src/openshiftCommonUI.module.js
```diff
@@ -22,6 +22,14 @@ angular.module('openshiftCommonUI', [])
   maxlength: 253,
   description: 'Name must consist of lower-case letters, numbers, periods, and hyphens. It must start and end with a letter or a number.'
 })
+// URI_PATH_FRAGMENT_VALIDATION patterns must consist of following characters:
+// a-z A-Z 0-9 . - _ ~ ! $ & ' ( ) * + , ; = : @
+// https://stackoverflow.com/questions/4669692/valid-characters-for-directory-part-of-a-url-for-short-links
+.constant('URI_PATH_FRAGMENT_VALIDATION', {
+  pattern: /^[a-zA-Z0-9.\-_~!$&'()*+,;=:@]+$/,
```
spadgett: I still think we might want to be more restrictive. From the stack overflow post you link:

> Keep in mind, though, that many applications (not necessarily browsers) that attempt to parse URIs to make them clickable, for example, may support a much smaller set of characters. This is akin to parsing e-mail addresses where most attempts also don't catch all addresses allowed by the standard.

But I'm OK with this if @jwforres prefers.

src/openshiftCommonUI.module.js
```diff
@@ -22,6 +22,14 @@ angular.module('openshiftCommonUI', [])
...
+  minlength: 8,
```
spadgett: If this is a generic URI path validator and not specific to webhook secrets, it shouldn't have a minlength here. You would add that check specifically for webhook validation.

src/openshiftCommonUI.module.js
jwforres: this actually doesn't seem like a complete list, i know `%` is a valid path character because its used for encoded chars

we might be overcomplicating this.  URI.js will automatically encode a path that you add via the `.path("whatever")` function. You can then get the encoded path back out by calling path with no args `.path()` https://medialize.github.io/URI.js/docs.html#accessors-pathname

the validator could just run the value through that and if what comes back out is not equal to what went in, then the key contains characters that must be encoded and we should warn them.

spadgett: You can't have a bare `%`, though. It needs to be percent encoded because it is used for percent encoding :)

jhadvig:
Closing the PR after discussion with @jwforres and @spadgett to put the validation in the console repo since its webhook specific.

PR #285, opened 2018-01-23
[WIP] Add hpa annotations

benjaminapetersen:
HPA `v2beta1` requested from `v1` api will have these annotations.

@jwforres @spadgett

### DIFF
--- a/dist/origin-web-common-ui.js
+++ b/dist/origin-web-common-ui.js
@@ -1388,11 +1388,21 @@ angular.module('openshiftCommonUI')
     };
   })
   .filter('imageStreamTagIconClass', function(imageStreamTagAnnotationFilter) {
-  return function(resource, /* optional */ tagName) {
-    var icon = imageStreamTagAnnotationFilter(resource, "iconClass", tagName);
-    return (icon) ? icon : "fa fa-cube";
-  };
-});
+    return function(resource, /* optional */ tagName) {
+      var icon = imageStreamTagAnnotationFilter(resource, "iconClass", tagName);
+      return (icon) ? icon : "fa fa-cube";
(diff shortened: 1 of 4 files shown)


spadgett:
Is it really easier / less risky to read and write these annotations instead just using the `v2beta1` API? We could only continue to only support target CPU in our form as long as we don't drop the other values.

benjaminapetersen:
Is a great question.  ATM I'm just working out reading them to test for the API & notify the user to go edit the HPA via cli.  I think if we wanted to write them, we should def update to the next version instead.

dist/origin-web-common-ui.js
```diff
@@ -1386,11 +1388,21 @@ angular.module('openshiftCommonUI')
     };
   })
   .filter('imageStreamTagIconClass', function(imageStreamTagAnnotationFilter) {
-  return function(resource, /* optional */ tagName) {
-    var icon = imageStreamTagAnnotationFilter(resource, "iconClass", tagName);
-    return (icon) ? icon : "fa fa-cube";
-  };
-});
+    return function(resource, /* optional */ tagName) {
```
benjaminapetersen: All I'm using this for is https://github.com/openshift/origin-web-console/pull/2708/files#diff-04a314c1b6d2d3fc082233f2ee89fa0cR106, reading, not writing.

benjaminapetersen: @spadgett ping, objections?  I'm not setting anything, this is just used to get at the annotations.  I could inline a similar fn, but figured this is where it should live.

benjaminapetersen:
I'll close this out per the `origin-web-console` review comments.

PR #282, opened 2018-01-22
Allow serial login flows to co-exist

liggitt:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1537120

PR #287, opened 2018-01-23
[3.6] Allow serial login flows to co-exist

liggitt:
pick of https://github.com/openshift/origin-web-common/pull/282


spadgett:
Looks like we need to add lodash#~3.10.1 to the resolutions in bower.json

PR #292, opened 2018-01-26
Prefer apps/v1 replicasets

spadgett:
I need to check if there are API changes, but we should be using replicasets from the apps group rather than extensions, which is deprecated.

/assign @jwforres 
/cc @benjaminapetersen


jwforres:
change lgtm, we are just making sure there are no negative side effects before we merge

spadgett:
Confirmed there are no changes that would break us.

PR #295, opened 2018-01-30
Align delete project modal with PatternFly standard

rhamilto:
I missed this one in [my first pass](https://github.com/openshift/origin-web-console/pull/2735).  &lt;sarcasm&gt;Hooray for multiple repos!&lt;/sarcasm&gt;

screen shot 2018-01-30 at 2 22 17 pm


src/styles/_core.less
```diff
@@ -1,23 +1,3 @@
-.delete-resource-modal {
```
spadgett: hey at least we get to remove more unnecessary code

PR #301, opened 2018-02-09
Add utilities for comparing versions

spadgett:
Take 2. I had to revert the other change because of a problem with the previous semver library and bower.

/assign @benjaminapetersen

### DIFF
--- a/dist/origin-web-common.js
+++ b/dist/origin-web-common.js
@@ -5787,6 +5787,36 @@ angular.module('openshiftCommonServices')
 });
 ;'use strict';
 
+angular.module("openshiftCommonServices")
+  .service("VersionsService", function(){
+    var compareInteral = function(v1, v2, newestFirst) {
+      v1 = v1 || '';
+      v2 = v2 || '';
+      try {
+        // compareVersions will sort via semver and throw an error if one of
+        // the values is not a version string.
+        var result = window.compareVersions(v1, v2);
+        return newestFirst ? result * -1 : result;
+      } catch (e) {
+        // One of the values is not a version string. Fall back to string comparison.
+        // Numbers will be sorted higher by localeCompare.
+        return v1.localeCompare(v2);
+      }
+    };
+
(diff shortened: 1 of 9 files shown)


test/spec/services/versionsServiceSpec.js
```diff
@@ -0,0 +1,26 @@
...
+
+  var VersionsService;
+  beforeEach(function() {
+    inject(function(_VersionsService_) {
+      VersionsService = _VersionsService_;
+    });
+  });
+
+  var versions = ["10.1", "9.1", "10.0", "9.0", "8.0", "10.1.1", "0.10", "latest", "old"];
+
```
benjaminapetersen: +1 tests, woot!

dist/origin-web-common.js
```diff
@@ -5787,6 +5787,34 @@ angular.module('openshiftCommonServices')
...
+        // the values is not a version string.
+        var result = window.compareVersions(v1, v2);
+        return newestFirst ? result * -1 : result;
+      } catch (e) {
+        // One of the values is not a version string. Fall back to string comparison.
+        // Numbers will be sorted higher by localeCompare.
+        return v1.localeCompare(v2);
+      }
+    };
+
```
benjaminapetersen: Quick Q ~ What do we get out of `rcompare`, assuming it means `reverseCompare`, correct?

spadgett: Yeah, the intention is that you can pass it to Arrays.sort. If you have to pass a 3rd param, that doesn't work.

Let me refactor this a bit to make it so compare and rcompare both take just the v1 and v2 params.

spadgett:
@benjaminapetersen updated

dist/origin-web-common.js
benjaminapetersen: Nice, this looks good to me!

benjaminapetersen:
@spadgett lgtm, thanks!

PR #303, opened 2018-02-14
Update npm dependencies to match origin-web-console repo

benjaminapetersen:
Per earlier conversation with @jwforres, this should solve the 3 items we listed.

In addition:
- new dist generated since this bumps our `uglify` plugin
- had to eliminate the `space_colon` option from the `Gruntfile` as it is no longer supported & `grunt build` failed 
- note: our minify JS for `console` and `common` are a bit different in the `Gruntfile` due to `usemin`, but the options match now.
- `jasmine-core` must now be explicitly declared


spadgett:
It looks like `npm install` failed in travis. @benjaminapetersen can you take a look?

benjaminapetersen:
yup checking it out

benjaminapetersen:
Bumped nodejs to 6 in travis to match console, tests now pass.

spadgett:
Hm, `grunt build` failing for me now because it can't launch Chrome :(

PR #304, opened 2018-02-19
Avoid runtime errors when requested resource doesn't exist

spadgett:
Update DataService to tolerate requests for resources that were not
found during discovery. This will show an error notification and reject
the promise rather than resulting in a runtime error.

/assign @jwforres 
@deads2k FYI

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -1521,6 +1521,27 @@ angular.module('openshiftCommonServices')
     addQueuedNotifications();
   };
 
+  var unknownResourceError = function(resource, opts) {
+    var message;
+    if (!resource) {
+      message = 'Internal error: API resource not specified.';
+    } else {
+      // APIService ResourceGroupVersion objects implement toString()
+      message = "Unknown resource: " + resource.toString();
+    }
+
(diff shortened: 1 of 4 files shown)


dist/origin-web-common-services.js
```diff
@@ -1521,6 +1521,27 @@ angular.module('openshiftCommonServices')
     addQueuedNotifications();
   };
 
+  var unknownResourceError = function(resource, opts) {
+    var message;
+    if (!resource) {
+      message = 'Interal error: API resource not specified.';
```
jwforres: typo, Internal

jwforres:
so this covers all the XHR, what are you thinking for the websockets?

spadgett:
We list before we open the websocket, so the list will fail.

jwforres:
@spadgett what about log streaming

jwforres:
i guess we get the pod first

spadgett:
Updated `DataService.createStream()`

spadgett:
Rebased to pick up #303

jwforres:
sorry just noticed the change to uniqueKey, how is that going to impact the functions relying on it, are any of them going to blow up with a null key

PR #307, opened 2018-02-20
Always return false in canI checks where resource is not set

spadgett:
Stop gap fix for https://github.com/openshift/origin-web-console/issues/2723

To scale stateful sets correctly, we need to support subresources whose group is different from the parent, but that is too risky a change for 3.9. I think we should add this guard to `canI` regardless.

/assign @benjaminapetersen 
@jwforres FYI


spadgett:
Opened #308

PR #316, opened 2018-03-12
Remove hawtio dependency

TheRealJon:
**Breaking Changes** Remove hawtio-core dependency. A pared down version of hawtio-core is now bundled with the dist scripts. Once this is approved and published, catalog and console will need to be adjusted to properly consume the pluginLoader provided by this repo.


Gruntfile.js
```diff
@@ -36,25 +36,51 @@ module.exports = function (grunt) {
       },
       concat: {
         options: {
-          separator: ';'
+          separator: '\n'
```
benjaminapetersen: I think since we are post-processing/minifying our `js`, we want to keep this separator as a `;`, just in case.  I think its a guard against someone forgetting to end a file with a semi, though a `\n` may solve the same problem.

TheRealJon: Can we do something like ";\n"? When looking at the resulting minified files, at least one newline made it easier to see where one file ended and the next started. I was inspecting the results of the build and they were just a little difficult to read sometimes.

Gruntfile.js
```diff
@@ -36,25 +36,51 @@ module.exports = function (grunt) {
...
         },
         ui: {
-          src: ['src/**/*UI.module.js', 'dist/scripts/templates.js', 'src/components/**/*.js', 'src/filters/**/*.js', 'src/ui-services/**/*.js'],
+          src: [
+            'src/pluginLoader.js',
```
benjaminapetersen: Pondering whether `pluginLoader` fits better under `services` over `ui`, or if it should be in both.  We consume the `dist`, so I don't think it can break the console, loading it twice *probably* overwrites one version, but its hard to tell if this thing has any internal `caches` that would cause memory leaks.

TheRealJon: From what I could tell, these files are intended to be consumed individually and not together. If you only need UI components, you don't have to consume the Services file and vice versa, and if you need both you just consume the origin-web-common.js file.

test/karma.conf.js
```diff
@@ -28,19 +30,24 @@ module.exports = function(config) {
       "bower_components/uri.js/src/URITemplate.js",
       "bower_components/uri.js/src/jquery.URI.js",
       "bower_components/uri.js/src/URI.fragmentURI.js",
-      "bower_components/hawtio-core/dist/hawtio-core.js",
-      "bower_components/kubernetes-container-terminal/dist/container-terminal.js",
-      "bower_components/hawtio-extension-service/dist/hawtio-extension-service.js",
-      // load up the fixtures first
+
+      // load fixtures before source
```
benjaminapetersen: Did we mean to drop `kubernetes-container-terminal/dist/container-terminal.js`?

test/karma.conf.js
```diff
@@ -57,13 +64,13 @@ module.exports = function(config) {
     },
 
     // web server port
-    port: 9443,
+    port: 4443,
```
benjaminapetersen: port change needed?

test/karma.conf.js
TheRealJon: Yes, I ran this by Sam. It is not used in this repo. I think it was a copy/paste error at some point.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+
+        BootstrapPlugin._injector = angular.bootstrap(bootstrapEl, pluginLoader.getModules(), {
+          strictDi: false
+        });
+        log.debug("Bootstrapped application");
+      });
+    });
+    return BootstrapPlugin;
+})();
+
```
benjaminapetersen: So in the end, should we still name this thing `hawtioPluginLoader`, or leave it globally as `pluginLoader` or `openshiftPluginLoader` && manually update all the references?  Thinking on this a second.  I don't feel strongly either way atm.

test/karma.conf.js
TheRealJon: I think someone else changed the port in another PR, but it was 8443 before, which conflicted with the api server if it was running when you try to run tests.

benjaminapetersen (commented):
Just a few small comments, mainly need to know about that `container-terminal` script.

test/karma.conf.js
benjaminapetersen: OK great, good to drop it then.

Gruntfile.js
benjaminapetersen: Yup, that works for me.

src/pluginLoader.js
TheRealJon: This was a suggestion from Sam to avoid breaking downstream repos other than our own.

TheRealJon: My original plan was just to use the global pluginLoader variable and refactor in console and common. We can still do that, but the alias still might be necessary.

Gruntfile.js
TheRealJon: The other option would be to just leave it out of the concatenated files, then leave it to the consumer to load the pluginLoader script before the others.

src/pluginLoader.js
benjaminapetersen: Works for me then!  @spadgett advice tends to be spot on!

spadgett: Yeah, we documented `hawtioPluginLoader.addModule` as the way to add modules, so we have to keep that working to avoid breaking customer extensions. I'm OK renaming and fixing up our code as long as we keep it as an alias (with a comment explaining).

https://docs.openshift.org/latest/install_config/web_console_customization.html

TheRealJon: I'll add a comment to the LOC where I create the alias. I can also refactor to use 'pluginLoader' in our repos if we want to do that.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+ *
+ */
+
+// log initialization
+/* globals Logger window console document localStorage $ angular jQuery navigator Jolokia */
+(function() {
+  'use strict';
+
+  Logger.setLevel(Logger.INFO);
+  Logger.storagePrefix = 'hawtio';
```
spadgett: Hm, I don't see `storagePrefix` documented in js-logger or used anywhere below. Can we just remove?

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+    }
+    var logLevel = Logger.DEBUG;
+    try {
+      logLevel = JSON.parse(window.localStorage['logLevel']);
+    } catch (e) {
+      console.error("Failed to parse log level setting: ", e);
+    }
+    // console.log("Using log level: ", logLevel);
+    Logger.setLevel(logLevel);
+    if ('showLog' in window.localStorage) {
```
spadgett: We're not using `showLog` so this can be removed.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+  Logger.setLevel(Logger.INFO);
+  Logger.storagePrefix = 'hawtio';
+
+  Logger.oldGet = Logger.get;
+  Logger.loggers = {};
+  Logger.get = function(name) {
+    var answer = Logger.oldGet(name);
+    Logger.loggers[name] = answer;
+    return answer;
+  };
```
spadgett: I think we can remove lines 18-24. I don't see where `Logger.loggers` is ever used.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+        }
+      }
+    }
+    if ('logBuffer' in window.localStorage) {
+      var logBuffer = window.localStorage['logBuffer'];
+      window['LogBuffer'] = parseInt(logBuffer, 10);
+    } else {
+      window.localStorage['logBuffer'] = window['LogBuffer'];
+    }
+    if ('childLoggers' in window.localStorage) {
```
spadgett: Let's remove the `childLoggers` stuff. We're not using it.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+      }
+      if (hdlr && hdlr.apply) {
+        try {
+          hdlr.apply(MyConsole, messages);
+        } catch (e) {
+          MyConsole.log(messages);
+        }
+      }
+    };
+  }
```
spadgett: Let's delete lines 70-97 and use js-logger default handler to add the context name. (I know this is existing hawtio code.)

https://github.com/jonnyreeves/js-logger/blob/46f9df7b946d50574601bb394e5d39c9ebcea338/README.md#default-log-handler-function

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+    return obj && Logger.getType(obj) === 'Array';
+  };
+
+  Logger.isObject = function (obj) {
+    return obj && Logger.getType(obj) === 'Object';
+  };
+
+  Logger.isString = function(obj) {
+    return obj && Logger.getType(obj) === 'String';
+  };
```
spadgett: I think we can remove 99-118. I'd like to use vanilla js-logger as much as possible.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+      if (consoleLogger) {
+        consoleLogger(messages, context);
+      }
+      var interceptors = window['logInterceptors'];
+      for (var i = 0; i < interceptors.length; i++) {
+        interceptors[i](context.level.name, text);
+      }
+    }
+    onAdd();
+    postLog.forEach(function (func) { func(); });
```
spadgett: In fact, I think we can delete 120-250 also. Doesn't look like we're using any of it, and I'd rather not carry the code.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+  var bootstrapEl = document.documentElement;
+
+  self.log = log;
+
+  /**
+   * List of URLs that the plugin loader will try and discover
+   * plugins from
+   * @type {Array}
+   */
+  // self.urls = [];
```
spadgett: Let's remove this since it's commented out anyway.

spadgett (changes requested):
Thanks @TheRealJon 

I wish we could remove more from plugin loader, but I guess it's working and we should leave well enough alone.

src/pluginLoader.js
TheRealJon: I'll try removing and make sure nothing breaks.

src/pluginLoader.js
```diff
@@ -0,0 +1,660 @@
...
+        BootstrapPlugin._injector = angular.bootstrap(bootstrapEl, pluginLoader.getModules(), {
+          strictDi: false
+        });
+        log.debug("Bootstrapped application");
+      });
+    });
+    return BootstrapPlugin;
+})();
+
+var hawtioPluginLoader = pluginLoader;
```
spadgett: Let's make it explicitly `window.hawtioPluginLoader = pluginLoader` so we're sure it's always a global.

benjaminapetersen:
@spadgett had similar feelings... if it works, & we can drop the dependency, I wasn't sure if we wanted to gut it more or not.  I'd kind of like to not see `Hawtio` obscuring our `console.log` statements anymore.

spadgett (approved):

spadgett:
Thanks @TheRealJon, looks good. Is this ready? Do you mind squashing?

TheRealJon:
@spadgett Ready for merge.

PR #317, opened 2018-03-12
Bug 1449038 - Add a simple weighted search to KeywordService

spadgett:
https://trello.com/c/2fFSJMX1
https://bugzilla.redhat.com/show_bug.cgi?id=1449038

### DIFF
--- a/src/services/keyword.js
+++ b/src/services/keyword.js
@@ -1,7 +1,9 @@
 'use strict';
 
 angular.module("openshiftCommonServices")
-  .service("KeywordService", function(){
+  .service("KeywordService", function($filter) {
+
+    var displayName = $filter('displayName');
 
     var generateKeywords = function(filterText) {
       if (!filterText) {
@@ -46,8 +48,57 @@ angular.module("openshiftCommonServices")
       return filteredObjects;
     };
 
+    // Perform a simple weighted search.
+    // weightedSearch is like filterForKeywords, except each field has a weight
+    // and the result is a sorted array of matches. filterFields is an array of
+    // objects with keys `path` and `weight`.
+    var weightedSearch = function(objects, filterFields, keywords) {
+      if (_.isEmpty(keywords)) {
+        return [];
+      }
+
+      var results = [];
+      _.each(objects, function(object) {
+        // Keep a score for matches, weighted by field.
+        var score = 0;
+        _.each(keywords, function(regex) {
+          var matchesKeyword = false;
+          _.each(filterFields, function(field) {
+            var value = _.get(object, field.path);
+            if (!value) {
+              return;
+            }
+
+            if (regex.test(value)) {
+              // For each matching keyword, add the field weight to the score.
+              score += field.weight;
+              matchesKeyword = true;
+            }
+          });
+
+          if (!matchesKeyword) {
+            // We've missed a keyword. Set score to 0 and short circuit the loop.
+            score = 0;
+            return false;
+          }
+        });
+
+        if (score > 0) {
+          results.push({
+            object: object,
+            score: score
+          });
+        }
+      });
+
+      // Sort first by score, then by display name for items that have the same score.
+      var orderedResult = _.orderBy(results, ['score', displayName], ['desc', 'asc']);
+      return _.map(orderedResult, 'object');
+    };
+
     return {
       filterForKeywords: filterForKeywords,
+      weightedSearch: weightedSearch,
       generateKeywords: generateKeywords
     };
   });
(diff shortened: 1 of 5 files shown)


spadgett:
/assign @benjaminapetersen 

PTAL

benjaminapetersen:
reviewing

test/spec/services/keywordServiceSpec.js
```diff
@@ -107,4 +107,94 @@ describe('KeywordService', function() {
...
+          name: 'mongo-db',
+          annotations: {
+            'openshift.io/display-name': 'Mongo DB',
+            tags: 'db'
+          }
+        }
+      }
+    };
+
+    it('should honor field weights', function() {
```
benjaminapetersen: So is this the kind of configuration we might want to do once up front, via `$provider` in `app.js` `.cofig()`, or would we want to custom config weights & paths for every use?

test/spec/services/keywordServiceSpec.js
```diff
@@ -107,4 +107,94 @@ describe('KeywordService', function() {
...
+        weight: 10
+      }, {
+        path: 'metadata.annotations.tags',
+        weight: 2
+      }], keywords);
+      expect(result).toEqual([ mockData['mongo-db'], mockData['maria-db'] ]);
+    });
+
+    it('should match all keywords', function() {
+      var keywords = KeywordService.generateKeywords('maria db');
```
benjaminapetersen: Maybe unrelated, but wonder if `.generateKeywords` should take either `'string of multiple'` or `['array','of','multiple']`.

src/services/keyword.js
```diff
@@ -46,8 +48,57 @@ angular.module("openshiftCommonServices")
...
+        }
+      });
+
+      // Sort first by score, then by display name for items that have the same score.
+      var orderedResult = _.orderBy(results, ['score', displayName], ['desc', 'asc']);
+      return _.map(orderedResult, 'object');
+    };
+
     return {
       filterForKeywords: filterForKeywords,
```
benjaminapetersen: Should we have both a `search` and a `weightedSearch`?  or could `weightedSearch` be used w/o weights, just with `keyword` and `objects` (perhaps if paths was last & optional)?

test/spec/services/keywordServiceSpec.js
benjaminapetersen: ie, `config` some functions & cache them, so we don't have copy / paste repeats of the setup in multiple places.  I don't know how  many times we will use this search feature.

src/services/keyword.js
spadgett: filterForKeywords is basically weightedSearch without the weights

src/services/keyword.js
```diff
@@ -46,8 +48,57 @@ angular.module("openshiftCommonServices")
...
 
+    // Perform a simple weighted search.
+    // weightedSearch is like filterForKeywords, except each field has a weight
+    // and the result is a sorted array of matches. filterFields is an array of
+    // objects with keys `path` and `weight`.
+    var weightedSearch = function(objects, filterFields, keywords) {
+      if (_.isEmpty(keywords)) {
+        return [];
+      }
+
```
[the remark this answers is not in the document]
spadgett: Can you give some example code?

src/services/keyword.js
benjaminapetersen: I suppose naming symmetry. Prob not big deal.

src/services/keyword.js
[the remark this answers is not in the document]
benjaminapetersen: Outermost: 

```JavaScript 
var weightedSearch = function(objects, filterFields, keywords) {
  if(_.isEmpty(keywords)) {
    return [];
  }
  
  var results = _.reduce(objects, function(result, object) {

    // do scoring work .....
    
    if (score > 0) {
      results.push({
        object: object,
        score: score
      });
    }
    return result;
  }, []);

  // Sort first by score, then by display name for items that have the same score.
  var orderedResult = _.orderBy(results, ['score', displayName], ['desc', 'asc']);
  return _.map(orderedResult, 'object');
  
}
```
Could possibly do the `score` and `matchesKeyword` similarly, but def isn't a huge deal.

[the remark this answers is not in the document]
spadgett: I guess I'm not clear on the benefit.

Since `_.reduce` can change result to anything, you have to read the code carefully to see that it's just appending values to an array. I feel like `_.each` / `push` is easier to understand.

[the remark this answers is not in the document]
benjaminapetersen: Right, I prob should not have suggested `reduce` over like `_.compact(_.map(objects))`, its too general purpose.

Its prob just the imperative vs declarative discussion... I'd rather not see the "noise" of manually making the new array(s) when we have a number of good functions that do that for us.

That said, is def a nit, and not a huge issue 😄

PR #318, opened 2018-03-15
Change project deletion confirmation message

rhamilto:
screen shot 2018-03-15 at 2 30 44 pm

Supports https://github.com/openshift/origin-web-console/issues/2864

cc: @beanh66


src/components/delete-project/delete-project-modal.html
```diff
@@ -14,7 +14,7 @@ <h1 class="modal-title">Are you sure you want to delete the project '<strong>{{p
         undone</strong>.  Make sure this is something you really want to do!
       </p>
       <div ng-show="typeNameToConfirm">
-        <p>Type the name of the project to confirm.</p>
+        <p>Type <strong>{{project | displayName}}</strong> to confirm deletion.</p>
```
spadgett: I might suggest

```html
Type <strong>{{project.metadata.name}}</strong> to confirm deletion.
```

Either name or display name will work, but name is probably easier to type. Also two projects can have the same display name, whereas name is unambiguous.

rhamilto: Changed.

beanh66:
@rhamilto LGTM

PR #320, opened 2018-03-22
Add mobilelients to API_PREFERRED_VERSIONS

witmicko:
@spadgett could you please have a look at this. It would help us clean up our code a little bit


spadgett (approved):
lgtm, but we should wait for 3.11

PR #321, opened 2018-04-09
Migrate from Bower to Yarn

TheRealJon:
- Remove all bower related files and references
- Change references to bower_components to node_modules
- Move bower dependency definitions over to package.json
- Remove unused dependencies
- Change install and clean deps scripts to use yarn
- Update docs


bower.json
```diff
@@ -1,49 +0,0 @@
-{
```
spadgett: @david-martin @sub-mod I know you're both using origin-web-common libraries. Is it a problem if we drop bower support completely in favor of yarn?

david-martin: @spadgett We did a standalone proof of concept that used origin-web-common last year, however that was dropped in favour of UI changes in the console instead.
No issue with changing to yarn

TheRealJon:
Obsolete as this repo is now maintenance only.

PR #322, opened 2018-04-10
[WIP] Request v1beta1/PartialObjectMetadataList

spadgett:
The PartialObjectMetadataList has been promoted to beta in 3.10. Use
content negotiation to fallback to JSON if PartialObjectMetadataList is
not supported.

/assign @jwforres


spadgett:
> So the issues around fallback not working as expected are fixed in upstream k8s now?

Yes, fixed upstream in 1.9.

PR #334, opened 2018-04-17
Assign metadata on binding creation

sedroche:
As part of ongoing mobile work we would like to be able to configure the metadata applied to ServiceBindings on creation. The aim is to add labels or annotations to the ServiceBinding to be able identify and then work with that specific binding.

### DIFF
--- a/dist/origin-web-common-services.js
+++ b/dist/origin-web-common-services.js
@@ -1652,16 +1652,15 @@ angular.module("openshiftCommonServices")
       return secret;
     };
 
-    var makeBinding = function(serviceInstance, application, parametersSecretName) {
+    var makeBinding = function(serviceInstance, application, parametersSecretName, metadata) {
       var instanceName = serviceInstance.metadata.name;
 
+      metadata = _.assign({ generateName: instanceName + '-' }, metadata);
       var credentialSecretName = generateSecretName(serviceInstance.metadata.name + '-credentials-');
       var binding = {
         kind: 'ServiceBinding',
         apiVersion: 'servicecatalog.k8s.io/v1beta1',
-        metadata: {
-          generateName: instanceName + '-'
-        },
+        metadata: metadata,
         spec: {
           instanceRef: {
             name: instanceName
@@ -1793,13 +1792,13 @@ angular.module("openshiftCommonServices")
       // is specified, also create a pod preset for that application using its
       // `spec.selector`. `serviceClass` is required to determine if any
       // parameters need to be set when creating the binding.
-      bindService: function(serviceInstance, application, serviceClass, parameters) {
+      bindService: function(serviceInstance, application, serviceClass, parameters, metadata) {
         var parametersSecretName;
         if (!_.isEmpty(parameters)) {
           parametersSecretName = generateSecretName(serviceInstance.metadata.name + '-bind-parameters-');
         }
 
-        var newBinding = makeBinding(serviceInstance, application, parametersSecretName);
+        var newBinding = makeBinding(serviceInstance, application, parametersSecretName, metadata);
         var context = {
           namespace: serviceInstance.metadata.namespace
         };
--- a/dist/origin-web-common.js
+++ b/dist/origin-web-common.js
@@ -2970,16 +2970,15 @@ angular.module("openshiftCommonServices")
       return secret;
     };
 
-    var makeBinding = function(serviceInstance, application, parametersSecretName) {
+    var makeBinding = function(serviceInstance, application, parametersSecretName, metadata) {
       var instanceName = serviceInstance.metadata.name;
 
+      metadata = _.assign({ generateName: instanceName + '-' }, metadata);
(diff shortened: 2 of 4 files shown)


sedroche:
@spadgett Does this change look ok? Following up PR on this issue https://github.com/openshift/origin-web-console/issues/2957

spadgett:
> As part of ongoing mobile work we would like to be able to configure the metadata applied to ServiceBindings on creation. The aim is to add labels or annotations to the ServiceBinding to be able identify and then work with that specific binding.

It's not clear to me how this metadata is going to be set. Is this going to be code in the console repo that handles certain service classes specially. I really want to avoid those special cases if we can.

We're already looking at a general way to handle labels when provisioning.

cc @pmorie

sedroche:
@spadgett The idea is to just pass the metadata to the bind modal component like:

https://github.com/aerogear/origin-web-console/pull/40/files#diff-9b6377c25f28da45ed50c1d71ef2863aR22

and then pass it on to the bindService using the changes defined in this PR. There is no special logic for specific service classes

spadgett:
It looks like this approach will only work if you create the binding from the overview. The annotations won't get added if you create the binding from the service instance page.

sedroche:
@spadgett yes it will be optional to set this binding metadata and we are ok with the annotations only being set from our UI.

sedroche:
Just to add a specific use case that we are trying to solve with this change which may clarify if we are on the right approach here.

When we create a binding between a mobile client and a mobile service this means that our physical client should be able to communicate with that service. To do that we need an updated SDK configuration (JSON object containing details on how to connect to the service) to add to the physical client. This SDK configuration will be displayed in a new component.

This component will react to whether a binding between client and service exists or not. If it does we go and fetch the config data and display it as shown below.

client-config-component

And the binding is created using our own UI component also (using the existing binding modal creation component)

our-own-binding-ui

So the problem we need to solve is that we need a way to identify that a binding exists between a specific client and specific service. To do that we decided on adding annotations to the binding ([as shown here](https://github.com/openshift/origin-web-common/pull/334#issuecomment-389874103)) 
But we may also want to add labels, or name the binding. And we also want to reuse the existing binding creation modal.  So this should be optional and allow configuration of any metadata properties.

A lot of the questions are about this PR are how we intend to use this change and it would be a small change similar to this -> https://github.com/openshift/origin-web-console/compare/master...sedroche:add-bindingmeta-to-bindservice?expand=1
I can go ahead and create this PR and we can continue the questions about how we use it there?

The question is whether you think this PR is a valid piece of work in it's own right? Can we merge here and continue the question of how to use it in a new PR or are we going about this the wrong way iyo and that we should try another solution?

wdyt @spadgett ?

benjaminapetersen:
@openshift/team-ux-review

beanh66:
@benjaminapetersen Per UX feedback, I think @cshinn left some comments in this PR: https://github.com/openshift/origin-web-console/pull/3021

benjaminapetersen:
@beanh66 great, thx!  @sedroche FYI on above, I think that feedback may address some items here.

dist/origin-web-common-services.js
```diff
@@ -1652,16 +1652,15 @@ angular.module("openshiftCommonServices")
       return secret;
     };
 
-    var makeBinding = function(serviceInstance, application, parametersSecretName) {
+    var makeBinding = function(serviceInstance, application, parametersSecretName, metadata) {
       var instanceName = serviceInstance.metadata.name;
 
+      metadata = _.assign({ generateName: instanceName + '-' }, metadata);
```
benjaminapetersen, line 1658: Consider making a copy `angular.copy(metadata)`?

benjaminapetersen: Also, `metadata` for what?  Should probably call it `bindingMetadata`. 

However, we are passing in `serviceInstance` and `application`, full objects, seems strange not to just pass in a full `binding` object.

benjaminapetersen (changes requested):
I'm not sure the signature of `makeBinding` is quite right.  Lets review.

dist/origin-web-common-services.js
sedroche: > Consider making a copy angular.copy(metadata)?

I believe `_.assign` will copy the properties onto the new target object `{ generateName: instanceName + '-' }`

> Also, metadata for what? Should probably call it bindingMetadata

Since the function signature is `makeBinding` i think it's implied that `metadata` is going to be part of the binding but I can update for clarity if you think it needs it?

> However, we are passing in serviceInstance and application, full objects, seems strange not to just pass in a full binding object.

We are creating the binding here so the only extra config needed is `parametersSecretName, metadata`

benjaminapetersen: Right, sorry, just thinking out loud a bit.

That is true on the `makeBinding()` name, so prob no point in being excessively verbose.  I may have no issue here, actually.  

Thanks!

PR #340, opened 2018-06-18
Add mobile download annotations

sedroche:
https://github.com/openshift/origin-web-console/pull/3020#discussion_r195473962


benjaminapetersen:
Great. If this applies to a number of your other PR's we should be sure to reference it as a dependency & merge first.

PR #342, opened 2018-08-22
Check whether current user is operations user

josefkarasek:
BZ: Cluster admin, click "View Archive" shows no result in Kibana
https://bugzilla.redhat.com/show_bug.cgi?id=1587807

### DIFF
--- a/src/services/aggregatedLoggingService.js
+++ b/src/services/aggregatedLoggingService.js
@@ -1,0 +1,74 @@
+'use strict';
+
+angular.module("openshiftCommonServices")
+  .factory("AggregatedLoggingService", function($q, $cacheFactory, Logger, $interval, APIService, DataService) {
+
+    var currentUser = null;
+    var cachedUserPermissions = $cacheFactory('operationsUsersCache', {
+          number: 10
+        });
+
+    inFlightPermissionsRequests = {};
+
+    // Create SelfSubjectAccessReview request againt authorization API to check whether
+    // current user can view pods/log from 'openshift-logging' project.
+    // Users who can view such logs are 'operations' users
+    var fetchUserPermisions = function(userName) {
+      var deferred = $q.defer();
+      currentUser = userName;
+      var userAllowed = cachedUserPermissions.get(userName);
+
+      if (!userAllowed) {
+          // If a request is already in flight, return the promise for that request.
+          if (inFlightPermissionsRequests[userName]) {
+            return inFlightPermissionsRequests[userName];
+          }
+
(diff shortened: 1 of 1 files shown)


spadgett:
I'd just include this in your origin-web-console PR. We are feature frozen and have moved new development to origin/console, so this will likely only be used for the one logging fix.

src/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,74 @@
...
+          number: 10
+        });
+
+    inFlightPermissionsRequests = {};
+
+    // Create SelfSubjectAccessReview request againt authorization API to check whether
+    // current user can view pods/log from 'openshift-logging' project.
+    // Users who can view such logs are 'operations' users
+    var fetchUserPermisions = function(userName) {
+      var deferred = $q.defer();
```
spadgett: You shouldn't need to create a new deferred. Just return the promise from `DataService.create()` or return `$q.when(cachedValue)` when returning a cached value.

src/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+angular.module("openshiftCommonServices")
+  .factory("AggregatedLoggingService", function($q, $cacheFactory, Logger, $interval, APIService, DataService) {
+
+    var currentUser = null;
+    var cachedUserPermissions = $cacheFactory('operationsUsersCache', {
```
spadgett: This seems like overkill since we're only ever checking one permission. It's really just one boolean value and a promise that we need to track.

src/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,74 @@
...
+          number: 10
+        });
+
+    inFlightPermissionsRequests = {};
```
spadgett: There should only ever be one, so this doesn't need to be a map.

src/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,74 @@
...
+            delete inFlightPermissionsRequests[userName];
+          });
+
+      } else {
+        // Using cached data.
+        Logger.log("AggregatedLoggingService, using cached user " + userName);
+        if ((_.now() - userAllowed.cacheTimestamp) >= 600000) {
+          userAllowed.forceRefresh = true;
+        }
+        deferred.resolve();
```
spadgett: Why aren't we returning the value in any of these `deferred.resolve()` calls?

spadgett (changes requested):

josefkarasek:
This file is part of openshift/origin-web-console#3060

